PR #120, opened 2026-03-10
feat: add web voice notes and RTL support

sam-saffron-jarvis:
## What

- add a `/v1/transcribe` web endpoint that accepts multipart audio uploads and routes them through the existing transcription provider config
- add a WhatsApp-style voice note button to the web composer so mobile users can record, transcribe, and send voice messages
- auto-send the transcript when the composer is empty, or append it into the draft when there is already text/attachments
- tighten the mobile composer layout so it sits lower on iPhone while still respecting the unsafe area
- render Hebrew and other RTL text correctly in the composer and message bubbles while keeping code, paths, and tool args LTR

## Why

The web UI worked fine for typed chat, but on mobile it was missing the obvious thing: hold up the phone, talk, send. The voice-note flow makes the web client much more natural on iPhone, and the RTL fixes stop Hebrew from looking broken while doing it.

## Testing

- `go test ./...`
- `go build ./...`
- `node --check internal/serveui/static/app-core.js`
- `node --check internal/serveui/static/app-render.js`
- `node --check internal/serveui/static/app-stream.js`
- manual smoke test:
  - started `term-llm serve web`
  - verified `/ui/healthz`
  - verified live UI served the new voice button/status markup
  - verified live JS/CSS included the new voice/mobile code
  - verified `/v1/transcribe` returned a sane validation error for unsupported upload content
- manual deploy test:
  - swapped the web service to the patched binary
  - restarted `webui`
  - verified `/chat/healthz`

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -74,6 +74,7 @@ Examples:
 All HTTP routes are mounted under --base-path (default /ui):
   POST {base}/v1/responses
   POST {base}/v1/chat/completions
+  POST {base}/v1/transcribe
   GET  {base}/v1/models
   GET  {base}/healthz
   GET  {base}/                       (web UI)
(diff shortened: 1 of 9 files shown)


sam-saffron-jarvis:
Follow-up context from manual iPhone testing:

- the mobile composer was intentionally pushed lower after the first pass because the original safe-area padding was leaving obvious usable space at the bottom on iPhone
- Hebrew / RTL rendering was added as part of this PR so both the composer and message bubbles behave properly for Hebrew, while code / paths / tool args stay LTR

So the mobile spacing and RTL bits are not incidental cleanup — they came directly out of testing the voice-note flow on phone.

PR #222, opened 2026-03-23
FEATURE: allow copying text with ctrl-c when selected

ducks:
Hitting ctrl+c to copy text out of the chat window was causing the chat to quit. this will allow you to copy text if you have it selected instead of closing chat.


SamSaffron:
yeah fair, ctrl-c is so loaded, thanks @ducks

PR #226, opened 2026-03-24
Add continuous tick for UI responsiveness during idle

ducks:
I have noticed that sometimes the UI becomes stale after asking a question, and I must send a new message for the screen to update.

Adds a 100ms tick to keep the UI responsive even when not actively streaming, so key events are processed promptly.


SamSaffron:
what are the actual symptoms here, during idle I don't really expect anything to be happening.

ducks:
> what are the actual symptoms here

there have been times where the UI will not update/stream until I send another message even though it's done thinking and the spinner has stopped.

SamSaffron:
yeah I am noticing some of this too on latest builds ... going to try some fixes..

PR #247, opened 2026-03-29
fix: Reusing a session ID with a different provider silently keeps the old provider

sam-saffron-jarvis:
## What

- detect when a stateful serve session is reused with a different requested provider
- return an explicit error instead of silently reusing the existing runtime
- add a regression test covering a reused session ID with a provider override

## Why

Session runtimes are keyed by session ID, so reusing a session ID with a different non-default provider could previously return the existing runtime and ignore the requested provider. That could send requests to the wrong backend and mix conversation state across providers with no error.

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -881,6 +881,13 @@ func (s *serveServer) runtimeForProviderRequest(ctx context.Context, sessionID s
 	if err != nil {
 		return nil, false, err
 	}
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing: the current behavior is intentional. The UI provider button changes the provider for new conversations, not mid-session. Returning a 400 here would break that flow. A proper mid-conversation provider switch would require history handover (squashing tool calls/results, etc.) which is out of scope for a bug fix.

PR #257, opened 2026-03-29
fix: Arbitrary local files can be copied into the web-served images directory

sam-saffron-jarvis:
## What

- validate tool-reported image paths before serving or copying them into the web UI image directory
- reject non-image files instead of copying arbitrary local file contents into `/images/`
- add coverage for valid image copies and non-image rejection

## Why

`ensureImageServeable` previously trusted any path returned by a tool and copied it into the web-served image directory. That allowed a malicious or compromised tool/MCP server to exfiltrate readable local files by reporting them as tool images. Restricting this path to actual image content closes that file-copy exfiltration path while preserving normal image serving.

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -191,6 +191,10 @@ func (s *serveServer) ensureImageServeable(imgPath string) (string, bool) {
 		log.Printf("[serve] ensureImageServeable: abs(%s): %v", imgPath, err)
 		return "", false
 	}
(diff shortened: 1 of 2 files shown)


sam-saffron-jarvis:
Closing — the content-type check is a valid security concern but it breaks intentional arbitrary-file usage (videos, PDFs, etc in web chat). Replacing with a `--files-dir` approach: image serving stays restrictive by default, but a separate served directory accepts any file type for explicit opt-in.

PR #356, opened 2026-04-12
fix: preserve blockquote structure in tui

sam-saffron-jarvis:
## What

Improves TUI blockquote rendering so quote structure stays visible.

## Why

The renderer was prefixing blockquotes with plain spaces, which flattened quote structure and made nested quotes hard to read. This PR:

- adds visible `> ` quote prefixes
- preserves blank quoted separator lines between child blocks
- updates the blockquote golden output and adds a nested blockquote regression test

## Testing

- `go test ./...`


sam-saffron-jarvis:
Closing this stale PR. The underlying blockquote structure fix is already on main; this PR is obsolete.

PR #357, opened 2026-04-12
fix: preserve loose list spacing

sam-saffron-jarvis:
## What

Preserves loose-list spacing in the TUI markdown renderer.

## Why

Loose lists were rendering exactly like tight lists, which dropped the blank-line paragraph spacing implied by markdown like:

```markdown
- First item

- Second item

- Third item
```

This PR respects `list.IsTight` when joining list item bodies and list items, and adds a regression test covering the loose-list case.

## Testing

- `go test ./...`


sam-saffron-jarvis:
Closing this stale PR. The loose list spacing fix is already on main; this PR is obsolete.

PR #358, opened 2026-04-12
fix: use commonmark tab stops in renderer

sam-saffron-jarvis:
## What

Switches markdown tab normalization to CommonMark-style 4-column tab stops in the TUI renderer.

## Why

The renderer was deliberately replacing every tab with two spaces before parsing. That is not CommonMark-compatible and breaks alignment/indentation-sensitive markdown.

This PR:

- expands tabs to the next 4-column stop instead of a blind two-space replacement
- updates the fenced-code tab golden output
- adds a regression test showing that a leading tab now becomes an indented code block

## Testing

- `go test ./...`


sam-saffron-jarvis:
Closing this for now. We're skipping the separate tab/4-space indentation issue in this pass.

PR #359, opened 2026-04-12
fix: size thematic breaks to render width

sam-saffron-jarvis:
## What

Sizes thematic breaks to the effective render width in the TUI markdown renderer.

## Why

The renderer was emitting a hardcoded eight-character rule for all `---` / `***` thematic breaks, which looked stubby and ignored viewport width.

This switches the rule to use the effective render width and adds width-based regression tests.

## Testing

- `go test ./...`


sam-saffron-jarvis:
Closing this stale PR. The thematic break width fix is already on main; this PR is obsolete.

PR #387, opened 2026-04-17
fix: Resetting a Responses conversation can be undone by a late stream completion

sam-saffron-jarvis:
## What

Add generation-guarded synchronization around `ResponsesClient.LastResponseID` so `ResetConversation` cannot be undone by a late `response.completed` event from an older stream.

Also add a regression test that resets the conversation mid-stream and verifies the next request does not send a stale `previous_response_id`.

## Why

Without this, clearing or starting a new conversation while an older Responses API stream is still finishing can restore the old response ID after the reset. The next request then chains onto the wrong server-side conversation and corrupts conversation state across threads.

### DIFF
--- a/internal/llm/responses_api.go
+++ b/internal/llm/responses_api.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"net/http"
 	"strings"
+	"sync"
 )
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #390, opened 2026-04-17
fix: Persisted serve sessions omit the effective system prompt

sam-saffron-jarvis:
## What

Persist the effective serve runtime system prompt whenever it is injected into a run.

This updates the runtime snapshot builders so the injected system message is included in in-memory history, the initial persisted snapshot, and the final persisted session snapshot. It also adds a regression test covering persisted serve sessions.

## Why

Serve runs already prepend `rt.systemPrompt` to the request when history and input do not contain a system message, but persisted session snapshots previously omitted that injected message. After eviction, restart, export, or later prompt/config changes, the restored session could continue under different instructions than the ones that originally governed the conversation.


SamSaffron:

PR #391, opened 2026-04-17
fix: Partial streamed assistant output is lost on agentic-loop stream errors

sam-saffron-jarvis:
## What

Persist partial assistant output from the agentic loop before returning stream errors.

This adds a small error-path callback in `internal/llm/engine.go` so already-streamed assistant text/reasoning is saved when `stream.Recv()` fails or when the provider emits an `EventError` mid-turn. If tool calls were already collected, the partial assistant message is sent through the response callback; otherwise it goes through the turn callback, matching existing persistence behavior as closely as possible.

Also adds regression coverage for both a raw `Recv` error and an `EventError` after partial output.

## Why

Today users can see partial assistant output in the UI, but if the agentic stream errors before normal turn completion that partial message is never persisted. That makes resumes, reconnects, and stored transcripts diverge from what was actually shown.


SamSaffron:

PR #393, opened 2026-04-18
fix(web): avoid duplicate effort label in header

sam-saffron-jarvis:
## What

Stop the web UI header from rendering a duplicate reasoning-effort badge when the selected model name already includes the same effort tier.

Example: `gpt-5.4-medium · medium` now renders as just `gpt-5.4-medium`.

## Why

The header is meant to be a compact status line. Showing the same tier twice is noisy and looks broken.

## Tests

- `node internal/serveui/static/app_stream_test.js`
- `node internal/serveui/static/app_core_test.js`
- `go build ./...`
- `go test ./...`

### DIFF
--- a/internal/serveui/static/app-core.js
+++ b/internal/serveui/static/app-core.js
@@ -395,6 +395,15 @@ const escapeHTML = (str) => {
   return div.innerHTML;
 };
 
+const normalizeHeaderStatToken = (value) => String(value || '').trim().toLowerCase();
+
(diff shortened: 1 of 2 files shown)


SamSaffron:
Superseded by existing `splitHeaderModelEffort` logic on main (internal/serveui/static/app-core.js:398), which already strips the duplicate effort suffix from the model name. The header now renders `gpt-5.4-medium · medium` as `gpt-5.4 · medium` via the existing path, making this change redundant. Closing as obsolete.

PR #400, opened 2026-04-18
fix: Claude CLI conversation state survives resets and leaks old history into new chats

sam-saffron-jarvis:
## What

Add a `ResetConversation()` hook to `ClaudeBinProvider` that clears the persisted Claude CLI resume state (`sessionID` and `messagesSent`) so fresh conversations do not reuse the previous Claude session. Also add a regression test covering the reset behavior.

## Why

`Engine.ResetConversation()` already forwards resets to providers that implement a reset hook, but `ClaudeBinProvider` was not implementing one even though it caches resume state across turns. That let supposedly fresh conversations keep resuming the old Claude CLI session, skip re-sending the full prompt/system instructions, and leak stale context into new chats.

### DIFF
--- a/internal/llm/claude_bin.go
+++ b/internal/llm/claude_bin.go
@@ -204,6 +204,13 @@ func (p *ClaudeBinProvider) SetToolExecutor(executor func(ctx context.Context, n
 	}
 }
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #401, opened 2026-04-18
fix: Telegram history carryover can reload huge image payloads despite a tiny carryover budget

sam-saffron-jarvis:
## What

- sanitize restored Telegram carryover history before storing it in `sess.history`
- replace carried-over user image parts with the existing `[image uploaded]` placeholder text
- strip image payloads from carried-over tool results while preserving the tool result structure/IDs
- add a regression test covering restored user-image and tool-result history with large image payloads

## Why

`restoreHistoryFromDB` chose the carryover suffix using placeholder text, but then restored the original `llm.Message` objects. That let a tiny `[image uploaded]` placeholder sneak megabytes of base64 image data back into memory and subsequent model requests after a restart or idle reset.

This keeps the carryover budget aligned with what is actually restored, while preserving enough structure for follow-up turns.

Note: `go test ./...` still reports an existing unrelated failure in `internal/serveui` (`TestStaticAssetsSupportEffortDropdown`).

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -586,7 +586,7 @@ func (m *telegramSessionMgr) restoreHistoryFromDB(ctx context.Context, chatID in
 			all = append(all, msg.ToLLMMessage())
 		}
 
-		history := tailMessages(all, maxChars)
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #402, opened 2026-04-18
fix: Tool-reported file paths are blindly republished as web-accessible files

sam-saffron-jarvis:
## What

- stop `ensureImageServeable` from copying arbitrary tool-reported image paths into the web-served image directory
- tighten `ensureFileServeable` so it only republishes files already inside the configured files dir or the configured image output dir
- canonicalize paths with symlink resolution before validating containment, and add regression coverage for rejected external paths plus the allowed image-output-dir copy path

## Why

Tool results should not be able to turn an arbitrary readable host path into a `/files` or `/images` URL. This change closes that local-file disclosure path while preserving the intended serve flow for generated images that originate from the configured image output directory.

Note: `go build ./...` passes. `go test ./...` still hits a pre-existing unrelated failure in `internal/serveui` (`TestStaticAssetsSupportEffortDropdown`).

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -199,6 +199,46 @@ func serveRoutePath(baseRoute, baseDir, servedPath string) string {
 	return baseRoute + filepath.ToSlash(relPath)
 }
 
+func canonicalizeServeExistingPath(p string) (string, error) {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #403, opened 2026-04-18
fix: Claude multimodal support leaks temporary image files indefinitely

sam-saffron-jarvis:
## What

Track Claude CLI temp image files created from multimodal prompt/tool-result data and remove them when the turn or provider session finishes.

## Why

Without cleanup, every uploaded image or tool-result image materialized into `/tmp` is leaked indefinitely. Repeated multimodal use can eventually exhaust disk space and break later requests.


SamSaffron:

PR #404, opened 2026-04-18
fix: Stale OpenRouter cache triggers unbounded background refresh fan-out

sam-saffron-jarvis:
## What

- guard stale OpenRouter cache refreshes so only one background refresh goroutine can run at a time
- clear the in-flight flag when the refresh completes so later stale-cache reads can trigger a new refresh if needed
- add a regression test covering repeated stale-cache reads to ensure they fan out to a single background refresh request

## Why

When stale cached OpenRouter models were present, every call to `GetCachedOpenRouterModels` launched a background refresh and returned stale data immediately. Under repeated completion or UI polling traffic this could create many concurrent refreshes against the same API and cache file, wasting resources and making model-list behavior unstable.

`go build ./...` passes.

`go test ./...` currently fails due to an existing unrelated `internal/serveui` test (`TestStaticAssetsSupportEffortDropdown`) in this worktree.

### DIFF
--- a/internal/llm/openrouter_models.go
+++ b/internal/llm/openrouter_models.go
@@ -5,12 +5,15 @@ import (
 	"os"
 	"sort"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/samsaffron/term-llm/internal/cache"
 )
 
 const openRouterCacheKey = "openrouter"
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #442, opened 2026-04-23
fix: use errors.Is for EOF checks in readSSELine to handle wrapped errors

sam-saffron-jarvis:
## What

Use `errors.Is()` instead of `==` when checking for `io.EOF` and `io.ErrUnexpectedEOF` in `readSSELine` (`internal/llm/openai_compat.go`).

## Why

Go's `net/http` chunked transfer decoder can return `io.ErrUnexpectedEOF` wrapped inside another error (e.g. via `fmt.Errorf("...: %w", io.ErrUnexpectedEOF)`). The direct `==` comparison fails for wrapped errors, so the error escaped `readSSELine` as a non-nil `err` return and was formatted into:

```
Qwen36local streaming error: unexpected EOF
```

This caused the stream to abort with an error instead of being treated as a clean end-of-stream — even when the local model server had finished sending all its data and simply closed the connection.

## How

- Added `"errors"` import to `openai_compat.go`
- Changed `err == io.EOF || err == io.ErrUnexpectedEOF` → `errors.Is(err, io.EOF) || errors.Is(err, io.ErrUnexpectedEOF)`

All existing tests pass.

### DIFF
--- a/internal/llm/openai_compat.go
+++ b/internal/llm/openai_compat.go
@@ -331,7 +331,7 @@ func (p *OpenAICompatProvider) ListModels(ctx context.Context) ([]ModelInfo, err
 func readSSELine(reader *bufio.Reader) (line string, eof bool, err error) {
 	line, err = reader.ReadString('\n')
 	if err != nil {
-		if err == io.EOF {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #445, opened 2026-04-24
fix: use 127.0.0.1 instead of localhost for Ollama default URL

sam-saffron-jarvis:
## What

Replace `http://localhost:11434` with `http://127.0.0.1:11434` as the default Ollama base URL in three places:

- `internal/llm/ollama.go` — `ollamaChatDefaultBaseURL` constant
- `internal/config/config.go` — default value for `embed.ollama.base_url`
- `internal/embedding/provider.go` — fallback in the embedding provider factory

## Why

On Linux (and other dual-stack systems), Go's DNS resolver resolves `localhost` to the IPv6 address `::1` before falling back to `127.0.0.1`. Ollama's default listener binds to `0.0.0.0:11434` (IPv4 only on many installs), so connections to `[::1]:11434` are refused even when Ollama is running fine.

Live session error:
```
dial tcp [::1]:11434: connect: connection refused
```

Using the explicit IPv4 loopback `127.0.0.1` bypasses the IPv6 resolution entirely and connects directly to where Ollama is listening.

## How

Minimal constant/default change — no logic altered. All existing tests pass.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -401,7 +401,7 @@ type EmbedVoyageConfig struct {
 
 // EmbedOllamaConfig configures Ollama embedding generation
 type EmbedOllamaConfig struct {
-	BaseURL string `mapstructure:"base_url"` // default: http://localhost:11434
+	BaseURL string `mapstructure:"base_url"` // default: http://127.0.0.1:11434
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #448, opened 2026-04-24
fix: OpenAI-compatible SSE parser treats malformed/truncated final chunks as success

sam-saffron-jarvis:
## What

- treat invalid JSON `data:` chunks in the OpenAI-compatible chat/completions SSE stream as terminal errors instead of skipping or breaking successfully
- add a regression test covering a malformed final chunk after partial tool-call state has been accumulated

## Why

A truncated or malformed final SSE chunk could previously fall through as a successful completion, emit `EventDone`, and flush partial tool calls with arguments degraded to `{}`. Returning an error preserves partial output already received while preventing incomplete tool calls and false success on broken streams.


SamSaffron:

PR #449, opened 2026-04-24
fix: Responses API streams can silently execute truncated or malformed tool calls

sam-saffron-jarvis:
## What

- fail Responses API streaming as soon as a recognized SSE event cannot be decoded
- fail the stream if it ends with unfinished tool calls instead of synthesizing fallback IDs/`{}` arguments from partial state
- add regression coverage for malformed tool-call events and truncated tool-call streams

## Why

A truncated or malformed Responses API SSE stream could previously be treated as a successful completion. In tool-call flows that meant term-llm could emit incomplete assistant output and even execute incorrect tool calls built from partial state. These streams should fail loudly instead of silently continuing.


SamSaffron:

PR #450, opened 2026-04-24
fix: Responses API stream corruption can silently synthesize bogus tool calls

sam-saffron-jarvis:
## What

- fail the Responses API stream when tool-call SSE events cannot be decoded instead of silently skipping them
- validate accumulated streamed tool calls before emitting them so incomplete, missing-ID, missing-arguments, or invalid-JSON calls do not get synthesized and returned as success
- add regression coverage for malformed tool-call deltas and truncated streams that previously could emit bogus tool calls

## Why

A truncated or malformed Responses API stream could previously skip `response.output_item.done` or argument delta events, then still flush `toolState.Calls()` at the end with fallback IDs and `{}` arguments. That turned upstream stream corruption into incorrect tool execution and corrupted assistant output instead of surfacing a hard error to the caller.


SamSaffron:

PR #451, opened 2026-04-24
fix: Inline image uploads bypass the configured max attachment size limit

sam-saffron-jarvis:
## What

Reject inline image uploads when saving the original attachment fails, instead of logging and continuing.
Added a regression test covering oversized inline images that exceed the configured per-file attachment limit.

## Why

Inline native image parts were still decoded and embedded into the LLM request after `saveUploadedFile` rejected them for exceeding `maxAttachmentBytes`.
That bypassed the intended 20 MB limit and allowed oversized request bodies to reach providers, wasting memory and causing avoidable failures.


SamSaffron:

PR #453, opened 2026-04-25
fix: improve connection error message in OpenAICompatProvider with service URL hint

sam-saffron-jarvis:
## What

When an `OpenAICompatProvider` (used for custom providers like `qwen36ollama`) fails to connect, the error message was a generic:

```
Qwen36ollama API request failed: Post "http://127.0.0.1:11434/v1/chat/completions": dial tcp 127.0.0.1:11434: connect: connection refused
```

After this fix, connection errors include the service URL and a actionable hint:

```
Qwen36ollama API request failed (is the service running at http://127.0.0.1:11434/v1?): Post "http://...": dial tcp ...: connect: connection refused
```

## Why

The native `OllamaProvider` already emits `"Ollama request failed (is Ollama running at %s?)"` on connection failure. The `OpenAICompatProvider` is also commonly used to front local servers (Ollama's OpenAI-compat endpoint, LM Studio, etc.) but gave no such hint, making the error harder to diagnose.

The fix detects `*net.OpError` (network-level errors like connection refused, connection reset, host unreachable) and enriches the message with the configured service URL. Non-network errors fall through to the original format unchanged.

## How

In `internal/llm/openai_compat.go`, `Stream()` now wraps the connection error check using `errors.As(err, &net.OpError{})` — both `net` and `errors` were already imported. The service URL is taken from `chatURL` if set, otherwise `baseURL`.

### DIFF
--- a/internal/llm/openai_compat.go
+++ b/internal/llm/openai_compat.go
@@ -410,6 +410,14 @@ func (p *OpenAICompatProvider) Stream(ctx context.Context, req Request) (Stream,
 	// Make HTTP request synchronously - this allows retry wrapper to catch errors like 429
 	resp, err := p.makeChatRequest(ctx, chatReq)
 	if err != nil {
+		var netErr *net.OpError
+		if errors.As(err, &netErr) {
+			serviceURL := p.chatURL
+			if serviceURL == "" {
+				serviceURL = p.baseURL
+			}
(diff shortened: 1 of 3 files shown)


SamSaffron:
Closing for now — defer; revisit later.

PR #458, opened 2026-04-25
fix: Standard remote image_url inputs are silently dropped from API requests

sam-saffron-jarvis:
## What

- download standard `http(s)` `image_url` inputs instead of silently dropping them when parsing user message content
- keep the existing inline `data:` URL behavior unchanged
- add focused tests for both responses-style `input_image` parts and chat-completions-style `image_url` parts

## Why

`parseUserMessageContent` claimed to accept standard `image_url` content parts, but only handled `data:` URLs and `continue`d on normal remote URLs. That made standards-compliant multimodal requests appear to succeed while the model never received the image. Downloading the remote image restores compatibility for both request shapes and preserves the existing attachment/save-to-disk flow.

### DIFF
--- a/cmd/serve_protocol.go
+++ b/cmd/serve_protocol.go
@@ -23,6 +23,8 @@ type sessionInterruptRequest struct {
 	Message string `json:"message"`
 }
 
+var remoteImageURLHTTPClient = &http.Client{Timeout: 30 * time.Second}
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing for now — defer; revisit later.

PR #462, opened 2026-04-27
speed up memory store opens

sam-saffron-jarvis:
## What

Speed up `internal/memory.NewStore` for existing `memory.db` files by recording the memory schema state in SQLite `PRAGMA user_version` and taking a fast path when it is already current.

The first open of an existing database still runs the full idempotent schema/FTS initialization and then stores the version. Later opens skip the repeated multi-statement schema DDL and migration probe.

## Why

Memory commands, completions, mining/status flows, and agent insight injection all open the memory store. Before this change, each open re-ran the full schema block and duplicate-column migration check even when the database was already initialized.

## Evidence

Added `BenchmarkNewStoreExistingDB`:

- Before: ~356 µs/op, ~5.45 KB/op, 113 allocs/op
- After: ~260 µs/op, ~4.87 KB/op, 84 allocs/op

Representative benchmark runs after change:

```text
BenchmarkNewStoreExistingDB-32  4545  264146 ns/op  4914 B/op  84 allocs/op
BenchmarkNewStoreExistingDB-32  4532  262852 ns/op  4871 B/op  84 allocs/op
BenchmarkNewStoreExistingDB-32  4479  259747 ns/op  4866 B/op  84 allocs/op
BenchmarkNewStoreExistingDB-32  4581  262328 ns/op  4868 B/op  84 allocs/op
BenchmarkNewStoreExistingDB-32  4737  252823 ns/op  4866 B/op  84 allocs/op
```

## Verification

- `go test ./internal/memory -run 'TestStoreSchemaVersionRecorded|TestStoreFragmentCRUDAndSearch|TestInsightListAndSearch'`
- `go test ./internal/memory`
- `go test ./internal/memory -run '^$' -bench BenchmarkNewStoreExistingDB -benchmem -count=5`
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/memory/store.go
+++ b/internal/memory/store.go
@@ -243,7 +243,15 @@ func NewStore(cfg Config) (*Store, error) {
 	return &Store{db: db}, nil
 }
 
+const memorySchemaVersion = 1
+
 func initSchema(db *sql.DB) error {
+	if current, err := isSchemaCurrent(db); err != nil {
+		return err
+	} else if current {
+		return nil
+	}
+
 	if _, err := db.Exec(schema); err != nil {
 		return fmt.Errorf("initialize memory schema: %w", err)
 	}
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing for now. The measured benefit is only about 90–100µs per existing memory DB open, while introducing schema-versioning/migration discipline risk. This feels too small to justify the maintenance footgun, especially compared with likely larger wins elsewhere such as session store hot paths.

PR #464, opened 2026-04-27
fix: return promptly on parallel tool cancellation

sam-saffron-jarvis:
## What

Make parallel tool execution return promptly when its context is canceled instead of waiting for every tool goroutine to finish.

The result channel is already sized to the number of tool calls, so workers can publish their final result after the caller has returned without blocking. This lets cancellation/stream close stop waiting on context-ignoring tools, and also removes the extra goroutine that only waited on a WaitGroup to close the result channel.

## Why

Terminal responsiveness on cancel matters most when tools are slow or stuck. Previously the parallel path waited for all worker goroutines before returning, so a cancellation could still block behind any non-cooperative tool. The new collector selects on `ctx.Done()` while preserving ordered results when all tools finish normally.

## Evidence

- Added `TestExecuteToolCallsParallelReturnsOnContextCancel`, which starts two context-ignoring tools and verifies `executeToolCalls` returns `context.Canceled` promptly after cancellation.
- Benchmark before: `BenchmarkExecuteToolCallsParallelPreCanceledContext` ~2.14 ms/op, ~4.1 KB/op, 48 allocs/op.
- Benchmark after: `BenchmarkExecuteToolCallsParallelPreCanceledContext` ~356 ns/op, 392 B/op, 7 allocs/op.

## Verification

- `gofmt -w internal/llm/engine.go internal/llm/engine_cancel_test.go`
- `go test ./internal/llm -run TestExecuteToolCallsParallelReturnsOnContextCancel -count=1`
- `go test ./internal/llm`
- `go build ./...`
- `go test ./...`

### DIFF
--- a/internal/llm/engine.go
+++ b/internal/llm/engine.go
@@ -1512,8 +1512,13 @@ func buildAssistantMessageWithReasoningMetadata(text string, toolCalls []ToolCal
 // Note: When executing in parallel, EventToolExecStart/EventToolExecEnd events
 // are emitted from concurrent goroutines. While the channel is thread-safe, events
 // may arrive in non-deterministic order. Consumers should use ToolCallID to correlate
-// start/end events rather than relying on ordering.
+// start/end events rather than relying on ordering. If ctx is canceled during
+// parallel execution, it returns promptly without waiting for context-ignoring tools.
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by changes already landed on main. The core cancellation behavior is present in current main; this branch is now stale/conflicting after later engine changes.

PR #495, opened 2026-05-01
fix: Alt-screen streaming rebuilds and resets the entire viewport content on each update

sam-saffron-jarvis:
## What

- avoid rebuilding a fresh `historyContent + streamingContent` string on append-only alt-screen streaming updates
- reuse the existing viewport line slice and append only the new streaming tail via `SetContentLines`
- reset the incremental append cache whenever width/history/turn state changes, and add focused tests for append-vs-rewrite behavior

## Why

Alt-screen streaming was re-concatenating the full transcript and forcing the viewport to rebuild from scratch on every streamed text/tool-state update. That makes long chats feel janky because normal token streaming repeatedly pays full-content allocation/splitting costs. This keeps the fast path incremental for append-only streaming while still falling back to the full rebuild path when content is rewritten or embedded UI changes.

### DIFF
--- a/internal/tui/chat/chat.go
+++ b/internal/tui/chat/chat.go
@@ -216,9 +216,11 @@ type Model struct {
 		// persist after streaming ends. Cleared when a new prompt is sent.
 		completedStream string
 		// Content versioning to avoid expensive string comparisons
-		contentVersion      uint64 // Incremented when content changes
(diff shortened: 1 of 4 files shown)


SamSaffron:
Closing so I can redo this locally with a small cache-invalidation hardening change before reopening.

PR #505, opened 2026-05-02
fix: UI throws away the initial POST SSE stream and immediately opens a second /events stream

sam-saffron-jarvis:
## What

Keep consuming the initial `POST /v1/responses` SSE stream after a successful send instead of immediately cancelling it and reconnecting to `GET /v1/responses/{id}/events` from `after=0`.

Also update the active stream tracking with the header response ID so the existing POST stream remains the tracked live transport, and only fall back to the resumable `/events` endpoint if the initial stream ends before reaching a terminal event.

## Why

The UI server path always returns `x-response-id` before streaming, so the previous flow created a POST stream, aborted it, and then replayed the same response from the start over a second SSE connection for every message.

That doubled the streaming HTTP work, delayed first visible output, and forced unnecessary reconnect/replay handling in the browser. This change preserves the existing resumable recovery path without paying for a second stream unless it is actually needed.

### DIFF
--- a/internal/serveui/static/app-stream.js
+++ b/internal/serveui/static/app-stream.js
@@ -129,16 +129,6 @@ const parseSSEStream = async (stream, onEvent) => {
   }
 };
 
-const cancelResponseBody = async (response) => {
-  if (!response?.body || typeof response.body.cancel !== 'function') return;
-  try {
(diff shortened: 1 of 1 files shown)


SamSaffron:

PR #514, opened 2026-05-03
fix: Copilot refreshes its session token after only 5 minutes, adding an avoidable pre-request roundtrip

sam-saffron-jarvis:
## What

- reduce Copilot's session-token refresh lead time from 20 minutes to 1 minute
- refresh when the token is missing, has no tracked expiry, or is actually close to expiring
- add regression tests covering both the keep-existing-token and near-expiry refresh paths

## Why

Copilot session tokens last about 25 minutes, but the previous logic refreshed them after roughly 5 minutes by checking `expiry - 20m`. That added an unnecessary `copilot_internal/v2/token` roundtrip to many prompts, directly hurting request startup latency and TTFT even though the existing session token was still valid.


SamSaffron:

PR #516, opened 2026-05-03
fix: Gemini tool-enabled turns are fully non-streaming

sam-saffron-jarvis:
## What

Use Gemini's streaming API even when tools are enabled, and parse streamed chunks for visible text plus tool calls/thought signatures as they arrive.

## Why

The previous tool path switched to a blocking `GenerateContent` call, so any agentic Gemini turn lost incremental output entirely and only emitted text/tool events after full model latency. This restores time-to-first-token for tool-heavy turns while preserving Gemini thought-signature propagation for tool results.


SamSaffron:
Closing in favor of #515, which covers the same Gemini tool-streaming fix and has passing CI.

PR #517, opened 2026-05-03
fix: One transient WebSocket setup failure permanently disables the faster Responses transport

sam-saffron-jarvis:
## What

Stop permanently disabling Responses-over-WebSocket after a single setup fallback. If WebSocket setup fails for one turn, the client now falls back to HTTP/SSE for that turn only and retries WebSocket on the next turn.

Also add a regression test that simulates a transient WebSocket setup failure followed by a successful WebSocket turn, ensuring the faster transport is retried instead of being latched off for the rest of the conversation.

## Why

OpenAI and ChatGPT commonly run with `use_websocket` enabled because the WebSocket transport improves TTFT and, for ChatGPT, preserves connection-local continuation state. Previously, any transient connect/handshake/write failure set a sticky disable flag that was only cleared by `ResetConversation`, silently degrading the rest of the conversation to slower HTTP/SSE requests.

Retaining WebSocket eligibility across transient setup failures restores the intended fast path automatically without requiring the user to clear the conversation.

### DIFF
--- a/internal/llm/responses_api.go
+++ b/internal/llm/responses_api.go
@@ -40,7 +40,6 @@ type ResponsesClient struct {
 	WebSocketConnectTimeout time.Duration
 	WebSocketWriteTimeout   time.Duration
 	WebSocketIdleTimeout    time.Duration
-	websocketDisabled       bool
 	wsMu                    sync.Mutex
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #521, opened 2026-05-04
fix: Inline image uploads are decoded twice and always synchronously written to disk before the model call

sam-saffron-jarvis:
## What

- stop saving LLM-native inline image uploads to disk on the request path before the provider call
- decode inline image base64 once, reuse the decoded bytes for size validation and optional resize, and only re-encode if the resized payload changed
- keep the existing save-to-disk behavior for non-native images and file uploads, and update coverage to assert inline LLM images no longer create upload files

## Why

The old upload path decoded every inline image once to save it, then decoded the same base64 again to decide whether it needed resizing, and sometimes encoded it a third time. It also always performed a blocking disk write before the model request even for image types the providers already accept inline. That adds avoidable CPU, memory copying, and filesystem latency to large or multi-image requests. This change keeps the same user-visible behavior for real file uploads while removing the extra decode and synchronous disk write from the hot path for inline LLM images.

### DIFF
--- a/cmd/serve_protocol.go
+++ b/cmd/serve_protocol.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
 	"mime"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as duplicate/superseded by #520, which has green CI and more complete regression coverage for the same inline-image upload hot-path fix.

PR #522, opened 2026-05-04
fix: Web UI turns every attachment into base64 data URLs and keeps multiple full copies in memory

sam-saffron-jarvis:
## What

Stop the Web UI from eagerly base64-encoding selected attachments in the composer.

This keeps pending attachments as `File` objects plus lightweight preview URLs, then only materializes a data URL at send time. The sent user message now reuses the same attachment objects instead of cloning another full base64 copy, and image rendering accepts either a preview URL or a data URL. Clearing a draft composer also revokes any pending blob preview URLs.

Also add a regression test covering the lazy attachment path.

## Why

The old flow created multiple full in-memory copies of every attachment before upload even began:

1. `handleFiles` read each file into a base64 data URL immediately
2. that payload was stored in `state.attachments`
3. it was copied again into `userMessage.attachments`
4. and then copied again into the JSON request body

That made large or multi-image prompts slower and much more memory-hungry in the Web UI. This change defers the expensive conversion until send time and removes the extra persistent copies while preserving previews and resend/rebuild behavior.

### DIFF
--- a/internal/serveui/static/app-render.js
+++ b/internal/serveui/static/app-render.js
@@ -938,12 +938,13 @@ const createMessageNode = (message) => {
       const attDiv = document.createElement('div');
       attDiv.className = 'message-attachments';
       message.attachments.forEach(att => {
-        if (att.type && att.type.startsWith('image/') && att.dataURL) {
+        const previewURL = att.previewURL || att.dataURL || '';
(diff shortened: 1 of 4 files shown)


SamSaffron:

PR #523, opened 2026-05-04
fix: Inline image uploads are decoded twice and synchronously written to disk before provider upload starts

sam-saffron-jarvis:
## What

Stop saving supported inline images to disk during `parseUserMessageContent`, and decode them only once in-memory before optional resize/compression for the provider payload.

## Why

Supported inline images already travel upstream as base64 image parts, so writing the original to disk on the request hot path was unnecessary work. This change removes blocking filesystem I/O for those uploads and avoids a second full base64 decode/allocation before image-generation or edit requests are sent.

### DIFF
--- a/cmd/serve_protocol.go
+++ b/cmd/serve_protocol.go
@@ -6,7 +6,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
-	"log"
 	"mime"
 	"net/http"
 	"os"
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as duplicate/superseded by #520, which has green CI and more complete regression coverage for the same inline-image upload hot-path fix.

PR #524, opened 2026-05-04
fix: Web attachment uploads inflate images into repeated base64 copies before the request even leaves the browser

sam-saffron-jarvis:
## What

- stop reading selected attachments into base64 data URLs at pick time
- keep pending web composer attachments as `File` objects plus blob preview URLs, and only materialize data URLs right before send
- reuse that one materialized payload for both the request body and the persisted user message, while revoking blob preview URLs when attachments are removed or cleared

## Why

The previous flow inflated every selected image immediately, then carried the same large payload through multiple intermediate structures before `fetch` even started uploading. Large or multi-image prompts could therefore create avoidable heap growth, sluggish composer behavior, and slower upload start times. Deferring base64 conversion until send removes that eager memory spike while keeping the existing request format and retry behavior intact.

### DIFF
--- a/internal/serveui/static/app-stream.js
+++ b/internal/serveui/static/app-stream.js
@@ -2437,9 +2437,10 @@ const renderAttachments = () => {
     const chip = document.createElement('div');
     chip.className = 'attachment-chip';
 
-    if (att.type.startsWith('image/')) {
+    const previewSrc = att.previewURL || att.dataURL;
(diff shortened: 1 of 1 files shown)


SamSaffron:
Closing as duplicate/superseded by #522 for the Web UI lazy-attachment work. #522 has broader cleanup and regression test coverage; it should be reviewed/fixed and merged instead.

PR #528, opened 2026-05-05
fix: read_file keeps scanning to EOF after hitting output limits

sam-saffron-jarvis:
## What

- stop `read_file` as soon as `streamLineNumberedRange` hits `MaxLines` or `MaxBytes`
- only include an exact total-line count in the truncation footer when EOF was actually reached
- add coverage that verifies truncation no longer consumes the entire reader

## Why

`read_file` kept scanning to EOF after it already had enough output, just so it could print an exact total line count in the truncation note. On large files this makes one of the hottest navigation tools feel hung and burns deadline budget on unnecessary I/O. Stopping immediately makes paged reads return promptly while preserving exact totals when they are already known.

### DIFF
--- a/internal/tools/read.go
+++ b/internal/tools/read.go
@@ -171,6 +171,7 @@ func streamLineNumberedRange(ctx context.Context, reader *bufio.Reader, startLin
 	writtenLines := 0
 	truncated := false
 	byteCapped := false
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by #530, which is already merged on main.

PR #529, opened 2026-05-05
fix: parallel tool execution launches an unbounded goroutine per tool call

sam-saffron-jarvis:
## What

Bound parallel tool execution in `Engine.executeToolCalls` with a small worker pool instead of launching one goroutine per tool call.

Also adds a regression test that verifies parallel tool execution stays concurrent while respecting the configured worker limit.

## Why

A single model response could previously fan out into an unbounded number of concurrent local tool executions, which can oversubscribe CPU, disk, and process limits and hurt overall throughput. Using a bounded worker pool preserves parallel execution while adding backpressure and preventing runaway goroutine/process fan-out.

### DIFF
--- a/internal/llm/engine.go
+++ b/internal/llm/engine.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"log/slog"
+	"runtime"
 	"sort"
 	"strings"
 	"sync"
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by the parallel tool cap merged in db938ea.

PR #531, opened 2026-05-05
fix: parallel tool execution has no global concurrency limit

sam-saffron-jarvis:
## What

Cap parallel tool execution in `Engine.executeToolCalls` with a small worker pool instead of launching one goroutine per tool call.

Also add a regression test that proves a response with more tool calls than the limit does not start more than the configured number at once.

## Why

With `ParallelToolCalls` enabled, a single model response could previously start an unbounded number of shells, greps, reads, or subagents immediately. That could oversubscribe CPU and disk, hurt TUI responsiveness, and reduce overall throughput due to contention.

A fixed concurrency cap keeps useful parallelism for common multi-tool turns while preventing one response from flooding the process.

### DIFF
--- a/internal/llm/engine.go
+++ b/internal/llm/engine.go
@@ -16,10 +16,11 @@ import (
 )
 
 const (
-	defaultMaxTurns       = 20
-	stopSearchToolHint    = "IMPORTANT: Do not call any tools. Use the information already retrieved and answer directly."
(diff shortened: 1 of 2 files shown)


SamSaffron:
Redoing this locally on main with default max turns raised to 50 and parallel tool execution capped at 20.

PR #545, opened 2026-05-07
perf: cache plain text stream checks

sam-saffron-jarvis:
## What

- Adds an incremental plain-text eligibility check for assistant stream tails.
- Reuses the previous plain-text scan result while streamed content grows by ordinary prose characters.
- Falls back to the full markdown/math scanner when punctuation or newlines could introduce markdown syntax.

## Why

During long plain-text responses, the web UI was rescanning the entire streamed assistant body every render just to decide whether it could keep using the cheap text-node append path. That is unnecessary for the common case where deltas are just words/spaces.

This keeps the conservative markdown behavior, but avoids repeated full scans for the boring case. Boring, in UI perf, is where the money is.

## Tests

- `node internal/serveui/static/markdown_streaming_test.js`
- `go test ./internal/serveui ./cmd`
- `go build ./... && go test ./...`

### DIFF
--- a/internal/serveui/static/app-render.js
+++ b/internal/serveui/static/app-render.js
@@ -606,7 +606,9 @@ const getOrCreateAssistantStreamState = (message, body) => {
       rendering: false,
       rafId: 0,
       timerId: 0,
-      lastRenderAt: 0
+      lastRenderAt: 0,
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #546, opened 2026-05-07
perf: cache gzipped UI assets

sam-saffron-jarvis:
## What

- Caches gzip-compressed embedded UI asset bodies by the uncompressed body ETag.
- Reuses cached compressed bytes for repeated `Accept-Encoding: gzip` requests.
- Keeps existing cache-control, ETag, conditional request, and HEAD behavior unchanged.

## Why

The UI assets are embedded and versioned, but the server recompressed compressible assets on every request. That is wasted CPU for hot assets like `app.css`, `app-*.js`, `sw.js`, and the manifest.

The cache key is the content hash already used for ETags, so asset updates naturally get fresh compressed bodies.

## Tests

- `go test ./cmd`
- `go build ./... && go test ./...`

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -49,6 +49,23 @@ func uiAssetETag(data []byte) string {
 	return `"` + hex.EncodeToString(sum[:]) + `"`
 }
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by merged #547, which covers gzipped UI asset caching along with ETags/rendered index caching.

PR #552, opened 2026-05-07
perf: omit app-webrtc.js from page load when WebRTC is disabled

sam-saffron-jarvis:
## What

Omit `app-webrtc.js` from the page and service worker when WebRTC is not enabled.

- Strip the `<script src="app-webrtc.js?v=...">` tag from the rendered `index.html`
- Remove `'./app-webrtc.js?v=...'` from the service worker's `SHELL_ASSETS` list

WebRTC-enabled deployments (`--webrtc` flag / `TERM_LLM_WEBRTC_*` env) are **unaffected** — the script tag is preserved and the SW still precaches the asset.

## Why

`app-webrtc.js` is 21 KB uncompressed (~6 KB gzip). It is an IIFE that checks `window.__WEBRTC_ENABLED__` at startup and exits immediately when the flag is absent. Without the flag the entire file is dead code: the browser fetches it, parses it, executes it — and nothing happens.

For the vast majority of deployments (no `--webrtc` flag), this saves:
- ~6 KB of network transfer per page load
- JS parse + execute time for 21 KB of unused code (~10–15 ms on mid-range devices)
- One network request during service worker install

## Tests

- `TestRenderIndexHTML_DropsWebRTCScriptWhenDisabled` — script tag absent when `webrtcHeadSnippet` is empty
- `TestRenderIndexHTML_KeepsWebRTCScriptWhenEnabled` — script tag present when snippet is set
- `TestDropScriptTagContaining` — helper removes correct tag, leaves others
- `TestDropScriptTagContaining_NeedleAbsent` — no-op when needle not found
- `TestDropSWShellAssetContaining` — removes webrtc entry from real rendered SW, other assets intact

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -182,7 +182,11 @@ func (s *serveServer) handleUI(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 	if assetName == "sw.js" {
-		serveEmbeddedUIBytes(w, r, serveui.RenderServiceWorker(), "text/javascript", "no-cache", true)
+		sw := serveui.RenderServiceWorker()
+		if s.webrtcHeadSnippet == "" {
+			sw = dropSWShellAssetContaining(sw, "app-webrtc.js")
+		}
+		serveEmbeddedUIBytes(w, r, sw, "text/javascript", "no-cache", true)
 		return
 	}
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closed in favor of a cleaner implementation that avoids parsing rendered HTML/JS and uses explicit serveui render options/placeholders instead.

PR #553, opened 2026-05-07
perf: replace O(n²) session.find loops with Map/Set lookups in status poll

sam-saffron-jarvis:
## What

Replace two O(n²) `Array.find` loops in the sidebar status poll hot path with O(n) `Map`/`Set` lookups.

**`app-render.js` — `updateSidebarStatus`:**
```js
// Before — O(n) find per entry × m entries = O(n·m) per poll tick
const local = state.sessions.find((session) => session.id === entry.id) || null;

// After — build Map once, then O(1) per entry
const localSessionMap = new Map(state.sessions.map((s) => [s.id, s]));
const local = localSessionMap.get(entry.id) || null;
```

**`app-sessions.js` — `pollSidebarStatus`:**
```js
// Before — O(n) find per entry × m entries
const hasUnknown = data.sessions.some(
  (entry) => !state.sessions.find((s) => s.id === entry.id)
);

// After — build Set once, then O(1) per entry
const localIds = new Set(state.sessions.map((s) => s.id));
const hasUnknown = data.sessions.some((entry) => !localIds.has(entry.id));
```

## Why

The sidebar status poll fires every **2 seconds** when any session has an active run. With 50+ sessions and 20+ status entries, the old code performed up to 1000+ comparisons per tick. The new code uses O(n + m) instead of O(n·m).

### DIFF
--- a/internal/serveui/static/app-render.js
+++ b/internal/serveui/static/app-render.js
@@ -1595,8 +1595,9 @@ const updateSidebarStatus = (statusSessions) => {
   if (!Array.isArray(statusSessions)) return false;
   let changed = false;
   let orderChanged = false;
+  const localSessionMap = new Map(state.sessions.map((s) => [s.id, s]));
   for (const entry of statusSessions) {
-    const local = state.sessions.find((session) => session.id === entry.id) || null;
+    const local = localSessionMap.get(entry.id) || null;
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #557, opened 2026-05-07
perf: add preload hints for large first-party scripts

sam-saffron-jarvis:
## What

Add `<link rel="preload" as="script">` hints in `<head>` for the four largest first-party JS files:

| File | Size (uncompressed) |
|------|-------------------|
| `app-stream.js` | 119 KB |
| `app-render.js` | 57 KB |
| `app-core.js` | 47 KB |
| `app-sessions.js` | 42 KB |

The preload `href` attributes go through the same `RenderIndexHTML` versioning table as the `<script src>` attributes, so both point to the identical `?v=<hash>` URL.

## Why

These scripts live at the end of `<body>`. Without preload hints the browser only discovers them after parsing ~19 KB of HTML body markup — after the render-blocking `app.css` has already stalled the render pipeline.

Explicit `<link rel="preload">` tags in `<head>` let the browser start fetching the large JS files in parallel with `app.css` as soon as `<head>` is parsed. Since the preload URL and script URL are identical, the browser reuses the single in-flight fetch — no double-download.

This reduces time-to-interactive on first/cold load (before the service worker has cached the assets).

## Notes

- A test asserts preload `href` matches `<script src>` to catch future versioning drift
- No execution order changes — preload is a hint, not a directive
- Complementary to PR #550 (gzip), PR #552 (skip WebRTC script), PR #555 (ETag caching)


SamSaffron:

PR #558, opened 2026-05-07
perf: cache message node in streamState to skip DOM lookups per stream chunk

sam-saffron-jarvis:
## What

In `enqueueAssistantStreamUpdate`, add a fast path that skips two DOM lookups when `streamState` already exists for the message.

**Before:** every SSE streaming chunk called:
1. `findMessageElement(message.id)` — `querySelector('[data-message-id="..."]')` scan over the entire messages container
2. `node.querySelector('.message-body')` — child query on the found node

Both happen even though `assistantStreamStates` already holds all the information needed after the first chunk.

**After:** on the first chunk, `streamState.node` is stored alongside the existing `streamState.body`. On all subsequent chunks, the fast path reads `existingState.node` directly — no DOM queries.

## Why

`enqueueAssistantStreamUpdate` is called synchronously on every SSE delta chunk, outside the rAF rate-limiter. For a response with hundreds of chunks, this was hundreds of unnecessary DOM traversals. The fix eliminates both queries from every chunk after the first with no semantic change.

### DIFF
--- a/internal/serveui/static/app-render.js
+++ b/internal/serveui/static/app-render.js
@@ -803,6 +803,16 @@ const enqueueAssistantStreamUpdate = (message) => {
     return;
   }
 
+  const existingState = assistantStreamStates.get(message.id);
+  if (existingState) {
+    existingState.latestContent = String(message.content || '');
+    existingState.dirty = true;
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #560, opened 2026-05-07
perf: cache rendered index HTML with sync.Once

sam-saffron-jarvis:
## What

Wrap `renderIndexHTML()` in `sync.Once` so the rendered HTML is computed once and reused for every subsequent request.

## Why

`renderIndexHTML()` rebuilds the same ~19 KB byte slice on every page navigation:
- Marshals `basePath`, asset version, sidebar config, and optional VAPID/WebRTC snippets into JSON
- Runs a table of ~12 `bytes.ReplaceAll` calls over the 19 KB template to inject versioned asset URLs

All inputs are fixed at server startup:
- `cfg.basePath`, `cfg.sidebarSessions` — set once from flags/config
- `serveui.AssetVersion()` — already uses its own `sync.Once`
- `cfgRef.Serve.WebPush.VAPIDPublicKey`, `webrtcHeadSnippet` — set at startup

The output is therefore identical on every call. Caching it eliminates repeated string manipulation and allocations on every page load.

## Notes

Follows the same `xOnce`/`xField` pattern already used by `responseRunsOnce`/`responseRuns` in `serveServer`.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -947,6 +947,8 @@ type serveServer struct {
 	sessionToResponse sync.Map                // session_id (string) → latest response_id (string)
 	responseRunsOnce  sync.Once
 	responseRuns      *responseRunManager
+	indexHTMLOnce     sync.Once
+	indexHTMLBytes    []byte
 	webrtcHeadSnippet string // injected into index.html <head>; empty when WebRTC disabled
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by merged #547, which includes rendered index HTML caching plus the related static asset caching work.

PR #562, opened 2026-05-07
perf: cache gzip-compressed static assets to avoid re-compression

sam-saffron-jarvis:
## What

Add `uiCachedGzip`: compress an embedded static asset once, store the result in a package-level `sync.Map` (keyed by SHA-256 of the raw bytes), and return the cached slice on all subsequent calls.

Replace the inline `gzip.NewWriter` block in `serveEmbeddedUIBytes` with a call to `uiCachedGzip`.

## Why

`serveEmbeddedUIBytes` previously re-compressed embedded static assets on every request that included `Accept-Encoding: gzip`. For a 119 KB file like `app-stream.js` at `BestCompression` this takes ~10–20 ms on typical hardware.

While a single-user deployment typically only cold-requests each asset once (browser caches with `immutable`; SW caches the rest), a shared/multi-user server (or a browser that cleared its cache) re-triggers compression for every new client. The cache eliminates this repeated work.

Embedded assets are constant for the lifetime of the process so the `sync.Map` never needs eviction.

## Notes

- Complementary to PR #550 (gzip for dynamic JSON responses, which uses `BestSpeed` and is not cached)
- Two tests added: pointer-equality check on repeated calls, and byte-equality check on two HTTP requests for the same asset

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -49,6 +49,25 @@ func uiAssetETag(data []byte) string {
 	return `"` + hex.EncodeToString(sum[:]) + `"`
 }
 
+// uiGzipCache stores gzip-compressed versions of static embedded assets,
+// keyed by SHA-256 of the raw bytes. Embedded assets are constant for the
+// lifetime of the process so the cache never needs eviction.
+var uiGzipCache sync.Map // [32]byte → []byte
+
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by merged #547, which covers gzip-compressed static asset caching in the broader embedded UI asset cache change.

PR #563, opened 2026-05-07
perf: cache RenderManifest and RenderServiceWorker with sync.Once

sam-saffron-jarvis:
## What

Cache the output of `RenderManifest()` and `RenderServiceWorker()` in `serveui` using `sync.Once`, following the same pattern as the existing `assetVersionOnce` cache for `AssetVersion()`.

## Why

Both functions are called on every request for `manifest.webmanifest` and `sw.js`. They perform:
- An embedded filesystem read (`StaticAsset`)
- One or more `bytes.ReplaceAll` passes over the file content
- A fresh heap allocation for the result

The output is purely a function of the embedded assets (which are fixed at compile time), so it is identical across all requests for the lifetime of the server. Computing it once and reusing the result removes redundant allocations and CPU work on every page load.

### DIFF
--- a/internal/serveui/embed.go
+++ b/internal/serveui/embed.go
@@ -19,6 +19,12 @@ var staticFiles embed.FS
 var (
 	assetVersionOnce sync.Once
 	assetVersion     string
+
+	manifestOnce  sync.Once
+	manifestBytes []byte
+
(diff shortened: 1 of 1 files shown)


SamSaffron:

PR #564, opened 2026-05-07
perf: incremental renderMessages — skip full DOM rebuild on same-session message append

sam-saffron-jarvis:
## What

`renderMessages` previously did `elements.messages.innerHTML = ''` followed by a full `createMessageNode` loop on every call — even when the active session hadn't changed and only new messages were appended.

This PR adds an incremental fast path: when `renderMessages` is called for the same session and the message list is a strict append of what was already rendered, it skips the teardown and only appends the new nodes.

Two module-level variables track state:
- `_lastRenderedSessionId` — the session ID of the most recent full or incremental render
- `_lastRenderedMessageIds` — the message IDs in render order

On each call the prefix of `session.messages` is compared against `_lastRenderedMessageIds`. If all previously-rendered IDs still match in order, only the tail is appended. Any other change (session switch, deletion, reorder, replacement) falls back to the full rebuild path.

## Why

`renderMessages` is called after every server sync that touches the active session. For a long session with 50+ messages, the old code would destroy and recreate all DOM nodes — including running `renderAssistantMarkdown` for every assistant message — even when only one new message arrived. The fast path makes this O(new messages) instead of O(all messages).

## Tests

Three new behavioral tests in `app_render_test.js`:
- Incremental append reuses the same node objects for existing messages
- Session switch forces a full rebuild
- Non-incremental change (deletion/reorder) forces a full rebuild

### DIFF
--- a/internal/serveui/static/app-render.js
+++ b/internal/serveui/static/app-render.js
@@ -1569,21 +1569,62 @@ const updateToolGroupNode = (message) => {
   }
 };
 
+let _lastRenderedSessionId = null;
+let _lastRenderedMessageIds = [];
+
 const renderMessages = (forceScroll = false) => {
   const session = ensureActiveSession();
   resetAssistantStreamRenders();
-  elements.messages.innerHTML = '';
 
-  if (!session || !session.messages.length) {
+  const sessionId = session ? session.id : null;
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #566, opened 2026-05-07
perf: fast path for response.output_text.delta SSE events

sam-saffron-jarvis:
## Summary

Two Go-side optimizations on the SSE streaming hot path, targeting the `response.output_text.delta` event that fires ~100x/sec during token streaming.

### 1. Eliminate `map[string]any` allocation in `appendTextDeltaEventLocked`

Previously, every text delta token required:
- Allocating `map[string]any{"output_index": ..., "delta": ev.Text}`
- Mutating it to add `"sequence_number"`
- `json.Marshal` of the whole map
- Passing the map to `applyRecoveryEventLocked` which re-reads `payload["delta"]`

Now a dedicated `appendTextDeltaEvent(outputIndex int, delta string)` method:
- Builds the JSON payload directly with `strconv.AppendInt` + `json.Marshal(delta)` into a single `[]byte`
- Inlines the recovery update (append delta to `recoveryMessages[idx].Content`) without re-parsing from a map
- Zero map allocations on the token hot path

### 2. Consolidate `writeStoredResponseEvent` from 3 → 1 `fmt.Fprintf` calls

Each SSE event replayed to a reconnecting client previously issued 3 separate `fmt.Fprintf` calls (one per SSE field). Now it's a single formatted write, cutting format-string parsing and potential write overhead by 3x for that path.

### Changes
- `cmd/serve_response_runs.go`: new `appendTextDeltaEvent` / `appendTextDeltaEventLocked` methods; simplified `writeStoredResponseEvent`; updated call site in `appendResponseRunEvent`

### DIFF
--- a/cmd/serve_response_runs.go
+++ b/cmd/serve_response_runs.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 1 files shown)


SamSaffron:

PR #569, opened 2026-05-07
fix: Smooth-buffered streaming still triggers one full Bubble Tea cycle per provider delta

sam-saffron-jarvis:
## What

- defer the next blocking stream read when a text delta has already scheduled a smooth rendering tick
- let `ui.SmoothTickMsg` resume provider ingestion after releasing buffered words
- update the chat perf test to cover the deferred-read behavior

## Why

The chat TUI buffered text for smooth 60fps rendering, but it still immediately re-armed `listenForStreamEvents()` after every text delta. That kept forcing one Bubble Tea update/view cycle per provider chunk, defeating the intended coalescing path and causing unnecessary render churn under chatty providers. Deferring the next read until the pending smooth tick fires restores the intended batching and improves streaming/input responsiveness.

### DIFF
--- a/internal/tui/chat/chat.go
+++ b/internal/tui/chat/chat.go
@@ -1629,11 +1629,15 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 
 		// Continue listening for more events unless we're done or got an error.
-		// Keep stream ingestion driven by provider output; smooth rendering is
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by merged #570, which fixes the smooth-buffer streaming Bubble Tea update-per-delta issue with clean CI.

PR #573, opened 2026-05-07
fix: Status-line token estimation rescans the whole conversation on the render hot path

sam-saffron-jarvis:
## What

Cache the streaming status-line context-token estimate in the chat model instead of recomputing it from the full conversation on every render.

- store the latest streaming context estimate alongside the in-progress streaming snapshot
- refresh that cached estimate only when the streaming snapshot changes (stream start, assistant snapshot updates, and completed turn/tool-result appends)
- have `statusLineUsageParts()` read the cached estimate while streaming, with the existing heuristic fallback preserved for non-streaming / unseeded cases
- add regression coverage for cached streaming estimates and cache refreshes

## Why

`renderStatusLine()` runs on the TUI render hot path, including spinner ticks, smooth ticks, wave ticks, and keypresses. During streaming it was calling `EstimateTokens(buildMessagesForContextEstimate())`, which copied and rescanned the entire conversation just to repaint the footer. In long sessions that turns every redraw into O(total transcript size) work and makes streaming/scroll/input feel progressively worse.

By shifting that work to the much less frequent snapshot-update path, the footer stays accurate while avoiding repeated whole-conversation rescans during rendering.

### DIFF
--- a/internal/tui/chat/chat.go
+++ b/internal/tui/chat/chat.go
@@ -75,6 +75,7 @@ type Model struct {
 	contextEstimateMu                sync.Mutex
 	streamingContextMessages         []llm.Message
 	streamingContextPendingAssistant bool
+	streamingContextTokenEstimate    int
 
 	// Streaming channels
 	streamChan <-chan ui.StreamEvent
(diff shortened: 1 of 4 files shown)


SamSaffron:
Closing as superseded by merged #572, which fixes the status-line token estimation render hot-path rescan with clean CI.

PR #574, opened 2026-05-07
fix: Chat smooth-buffer streaming still processes one Bubble Tea cycle per incoming stream event

sam-saffron-jarvis:
## What

- defer the next chat stream read when a text event has already scheduled a smooth render tick
- let `ui.SmoothTickMsg` resume the blocked stream read after the frame, matching the existing ask-mode behavior
- update the chat regression test to cover deferred stream reads instead of immediate re-arming

## Why

Chat buffered streamed text into `m.smoothBuffer`, but still re-armed `listenForStreamEvents()` after every non-terminal event. Under chatty providers that meant each token chunk still forced a full Bubble Tea update/view cycle and kept the stream channel hot, which caused redraw storms, extra backpressure, and worse typing/scrolling smoothness. Deferring the next read until the pending smooth tick coalesces those render passes without broad refactoring.

### DIFF
--- a/internal/tui/chat/chat.go
+++ b/internal/tui/chat/chat.go
@@ -1629,11 +1629,15 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 
 		// Continue listening for more events unless we're done or got an error.
-		// Keep stream ingestion driven by provider output; smooth rendering is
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by merged #570, which fixes the smooth-buffer streaming Bubble Tea update-per-delta issue with clean CI.

PR #583, opened 2026-05-07
perf: eliminate String coercion and trim in isSessionVisible hot path

sam-saffron-jarvis:
## Summary

`isSessionVisible` is called once per streaming token (~100×/sec) to guard DOM updates to the currently visible session.

The previous implementation allocated a trimmed string on every call:

```js
// before
const isSessionVisible = (session) => {
  const sessionId = String(session?.id || '').trim();
  return Boolean(sessionId && !state.draftSessionActive && state.activeSessionId === sessionId);
};
```

Session IDs are always clean strings (generated by `generateId()`, no whitespace) and `state.activeSessionId` is always assigned directly from `session.id`. The defensive `String().trim()` has never been necessary.

```js
// after
const isSessionVisible = (session) => !state.draftSessionActive && state.activeSessionId === session.id;
```

## Impact

Eliminates ~100 unnecessary string allocations per second during active streaming: the `String()` coercion, `.trim()` allocation, `Boolean()` wrapper, and optional chaining overhead are all removed.

### DIFF
--- a/internal/serveui/static/app-stream.js
+++ b/internal/serveui/static/app-stream.js
@@ -377,10 +377,7 @@ const updateResponseSequence = (session, payload) => {
 // Visible-session render guards: stream events may arrive after the user has
 // switched discussions, so session data can update while DOM work is limited to
 // the conversation currently mounted in elements.messages.
-const isSessionVisible = (session) => {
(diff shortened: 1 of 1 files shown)


SamSaffron:

PR #599, opened 2026-05-08
fix: Skills initialization does a full multi-path skill catalog scan before every first prompt

sam-saffron-jarvis:
## What

- make `internal/skills.NewSetup` stop calling `registry.List()` eagerly during setup
- add a fast `Registry.HasAnySkill()` probe so startup only confirms that skills exist before enabling the system
- lazily generate `<available_skills>` metadata via `Setup.EnsurePromptMetadata()` when prompt injection actually needs it
- inject skill metadata before registering tools in ask/chat so `HasOverflow` is populated without restoring the eager scan path
- add a regression test covering lazy prompt metadata generation

## Why

`NewSetup` previously did a full multi-path skill catalog scan up front, parsing metadata for every discovered skill before the first prompt just to build `<available_skills>` XML. That adds avoidable startup latency to normal ask/chat/serve flows, especially when multiple project and ecosystem skill directories are present. Deferring the full catalog walk until prompt metadata is actually needed keeps startup lighter while preserving existing skill activation behavior.

### DIFF
--- a/cmd/ask.go
+++ b/cmd/ask.go
@@ -326,6 +326,7 @@ func runAsk(cmd *cobra.Command, args []string) error {
 	}
 	sessionID = ensureRequestSessionID(sessionID, resuming)
 	settings.SessionID = sessionID
+	settings.SystemPrompt = InjectSkillsMetadata(settings.SystemPrompt, skillsSetup)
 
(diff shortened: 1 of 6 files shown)


SamSaffron:

PR #603, opened 2026-05-08
perf: cache tool registry specs

sam-saffron-jarvis:
## What

Cache `ToolRegistry.AllSpecs()` results in deterministic name order and invalidate the cache when tools are registered or unregistered.

This keeps the returned slice/top-level `ToolSpec` values independent while documenting the existing immutability contract for `ToolSpec.Schema` maps.

## Why

`AllSpecs()` is called on request-building paths and in local test loops. The previous implementation rebuilt the tool-name slice, sorted it, and called every tool's `Spec()` on every call even when the registry was unchanged.

## Verification

- `go test -run '^$' -bench '^BenchmarkToolRegistryAllSpecs$' -benchmem -count=5 ./internal/llm`
- `go test -count=3 -run '^(TestToolRegistry(AllSpecsSortedByName|AllSpecsReturnedSliceIsIndependent|AllSpecsInvalidatesCache|ConcurrentAccess)|TestToolSpecsForRequestPreservesSortedOrderAfterFiltering)$' -v ./internal/llm`
- `go test ./internal/llm`
- `go build ./...`
- `go test -count=1 ./...`

## Evidence

Representative benchmark (`BenchmarkToolRegistryAllSpecs`, 128 registered tools):

- Before: ~20.3 µs/op, 51.8 KB/op, 258 allocs/op
- After: ~1.0 µs/op, 6.5 KB/op, 1 alloc/op

The concurrent registry stress test also dropped from ~1.29-1.35s per run to ~0.12-0.14s per run in focused `-count=3` runs.

### DIFF
--- a/internal/llm/tools.go
+++ b/internal/llm/tools.go
@@ -40,18 +40,23 @@ type FinishingTool interface {
 
 // ToolRegistry stores tools by name for execution.
 type ToolRegistry struct {
-	mu    sync.RWMutex
-	tools map[string]Tool
+	mu         sync.RWMutex
+	tools      map[string]Tool
+	specsCache []ToolSpec
+	specsDirty bool
 }
 
 func NewToolRegistry() *ToolRegistry {
-	return &ToolRegistry{tools: make(map[string]Tool)}
+	return &ToolRegistry{tools: make(map[string]Tool), specsDirty: true}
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #605, opened 2026-05-09
fix: Switching to a lazy-loaded session fetches the same message history twice

sam-saffron-jarvis:
## What

- track when `switchToSession()` has already successfully preloaded messages for a lazy-loaded (`_serverOnly`) session
- pass `skipMessagesFetch` into `syncActiveSessionFromServer()` only when that preload succeeded, so the session switch path avoids a second full-history GET while still preserving the sync fallback if preload fails
- add a regression test covering lazy-loaded session switching to assert only one `/messages` fetch happens

## Why

Switching to a lazy-loaded session was fetching the same message history twice back-to-back on the idle-session path: once during the explicit preload and again during runtime-state sync. That doubled session-switch latency and repeated the merge/render work right in a user-visible interaction. This change removes the redundant fetch without weakening recovery behavior.

### DIFF
--- a/internal/serveui/static/app-sessions.js
+++ b/internal/serveui/static/app-sessions.js
@@ -211,10 +211,12 @@ const switchToSession = async (sessionId, options = {}) => {
   updateURL(sessionSlug(session));
   refreshPendingInterjectionBanner();
 
+  let didPreloadServerMessages = false;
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #606, opened 2026-05-09
fix: /v1/models has no cache and can block initial /chat startup on upstream provider model listing

sam-saffron-jarvis:
## What

- make `/v1/models` prefer already-available local model lists (`providers.*.models`, configured default model, curated built-in lists) before attempting upstream `ListModels`
- add a small in-memory cache for upstream `ListModels` responses when `/v1/models` does need to fetch dynamically
- add regression tests covering the fast local path and the new memoized upstream path

## Why

The web UI waits on `/v1/models` during startup. Before this change, the handler could spend up to 15 seconds blocking on an upstream model-list API call even though `/v1/providers` already exposed enough model data for the UI to render immediately. This makes first-load latency provider-dependent and especially painful for slow model-list endpoints.

Serving the local/configured model list first removes that startup round trip in the common case, while memoizing dynamic listings avoids repeating the same slow upstream request for providers that truly need it.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -944,8 +944,9 @@ type serveServer struct {
 	shutdownOnce      sync.Once
 	modelsMu          sync.Mutex
 	modelsProviders   map[string]llm.Provider // keyed by provider name
-	responseToSession sync.Map                // response_id (string) → session_id (string)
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #607, opened 2026-05-09
fix: Session history loads the entire transcript, inlines image blobs, and returns it as one giant response

sam-saffron-jarvis:
## What

- paginate `/v1/sessions/{id}/messages` with a bounded page size and `has_more` / `next_offset` metadata instead of returning one unbounded response
- materialize stored inline image blobs into served `/images/` files so session history JSON no longer embeds large `data:` URLs
- preserve conditional reloads by hashing the response body together with the session `updated_at`, and teach the web UI to follow paginated history pages when needed
- add Go and serve-ui tests covering pagination and served image history payloads

## Why

Large transcripts were being loaded as one giant JSON payload, including base64 image blobs, whenever the web UI opened or reconciled a session. That scaled browser/server work with the full transcript size, caused avoidable memory churn, and made large chats noticeably slower to open and refresh. Paging the history and serving images out-of-band keeps each response bounded while preserving full session history and image rendering.

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"crypto/sha256"
 	"crypto/subtle"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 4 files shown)


SamSaffron:

PR #609, opened 2026-05-09
perf: amortize response run replay compaction

sam-saffron-jarvis:
## What

Amortize response run replay-window compaction by tracking a logical `eventStart` offset and reclaiming the dropped prefix in batches instead of copying the entire retained replay window on every streamed event after the replay limit.

This keeps reconnect replay semantics intact by iterating only `events[eventStart:]`, and adds:

- a regression test for replay order, `minReplayAfter`, bounded storage, and encoded payload sequence numbers
- a benchmark for the retained replay append path

## Why

Long streamed responses retain only the last 2048 SSE events for reconnect replay. Once the old slice exceeded that limit, every new text delta copied the full 2048-event replay window while holding the response run mutex. A CPU profile of the new benchmark on the baseline showed `compactEventsLocked`/`runtime.memmove` dominating the append path.

## Evidence

Baseline, before this change:

```text
go test ./cmd -run '^$' -bench '^BenchmarkResponseRunAppendTextDeltaRetainedReplay$' -benchmem -count=5
BenchmarkResponseRunAppendTextDeltaRetainedReplay-32  ~2241-2252 ns/op  80 B/op  1 allocs/op
```

Baseline CPU profile for that benchmark:

```text
runtime.memmove                                      96.60%
github.com/samsaffron/term-llm/cmd.(*responseRun).compactEventsLocked  97.28% cum
```

After this change:

```text
go test ./cmd -run '^$' -bench '^BenchmarkResponseRunAppendTextDeltaRetainedReplay$' -benchmem -count=1
BenchmarkResponseRunAppendTextDeltaRetainedReplay-32  21929305  52.85 ns/op  80 B/op  1 allocs/op
```

Also verified:

```text
go test ./...
go build ./...
```

### DIFF
--- a/cmd/serve_response_runs.go
+++ b/cmd/serve_response_runs.go
@@ -64,8 +64,10 @@ type responseRun struct {
 	usage              llm.Usage
 	sessionUsage       llm.Usage
 	lastSequenceNumber int64
-	// Retain the raw event stream until the run expires so reconnecting clients can replay by sequence number.
+	// events[eventStart:] is the retained replay window; dropped prefix slots
+	// are zeroed and reclaimed in batches to avoid per-token slice copies.
 	events             []responseRunEvent
+	eventStart         int
 	minReplayAfter     int64
 	maxRetainedEvents  int
 	recoveryMessages   []responseRunRecoveryMessage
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #611, opened 2026-05-10
fix: Tool-history sanitization deep-clones the entire conversation on every provider call

sam-saffron-jarvis:
## What

- add a fast path in `sanitizeToolHistory` that returns already-valid histories without deep-cloning every message part
- only fall back to the existing clone-and-rewrite logic when event rows must be filtered or tool call/result history actually needs repair
- add tests covering both the zero-clone fast path and orphaned tool-call repair behavior

## Why

`sanitizeToolHistory` runs in multiple hot request-preparation paths before provider calls. Deep-cloning every `ToolCall`, `ToolResult`, raw args blob, diff, image, and thought signature on every request adds avoidable allocation churn and startup latency for long sessions. This change keeps the existing sanitization behavior for invalid histories while making the common valid-history case allocation-light.


SamSaffron:

PR #612, opened 2026-05-10
fix: OpenAI/Responses turns rebuild the full transcript before applying previous_response_id or WS continuation

sam-saffron-jarvis:
## What

- stop `OpenAIProvider.Stream` from eagerly materializing `ResponsesRequest.Input` on every turn
- let `ResponsesClient` build OpenAI Responses input lazily from raw `[]Message` values
- add a continuation-only builder for latest user/tool follow-up payloads so `previous_response_id` and Responses-over-WebSocket continuation can send the incremental turn without paying a full-transcript walk first
- keep full-history reconstruction available for debug logging, 404 full-history retries, and post-stream websocket state tracking
- add regression coverage for continuation payload selection from raw messages

## Why

OpenAI/Responses turns were rebuilding and sanitizing the entire transcript before the shared client decided whether it could continue with `previous_response_id` or websocket state. That meant long conversations still paid the full client-side request-prep cost on every turn, which hurts time-to-first-token even when server-side continuation is available.

### DIFF
--- a/internal/llm/openai.go
+++ b/internal/llm/openai.go
@@ -126,7 +126,7 @@ func (p *OpenAIProvider) Stream(ctx context.Context, req Request) (Stream, error
 
 	responsesReq := ResponsesRequest{
 		Model:          model,
-		Input:          BuildResponsesInput(req.Messages),
+		Messages:       req.Messages,
 		Tools:          tools,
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #614, opened 2026-05-10
perf: cache skill registry metadata

sam-saffron-jarvis:
## What

Cache metadata-only `skills.Registry.List()` results behind a lightweight fingerprint of the configured skill search paths and each discovered `SKILL.md` file's mtime/size.

Also:
- reuse metadata discovered during prompt/search listing for later `Get()` activation, including frontmatter-name aliases
- fall back to normal search-path resolution if cached metadata points at a stale path
- return defensive copies from cached lists so callers cannot mutate registry cache state
- add tests for stale-cache fallback, defensive copies, and cache refresh on added skills
- add `BenchmarkRegistryListRepeated`

## Why

Chat/serve startup injects skill metadata, and long-running sessions can call `search_skills` repeatedly. Before this change each `List()` reparsed every `SKILL.md` frontmatter even when the registry contents had not changed. The new cache keeps the cheap filesystem freshness check but avoids repeated YAML parsing and allocation-heavy metadata reconstruction.

## Evidence

Benchmark with 200 temp skills, repeated metadata listing after warmup:

```text
# before (temp checkout of HEAD + benchmark)
BenchmarkRegistryListRepeated-32   513-606 iters   2.09-2.23 ms/op   ~2.87-2.89 MB/op   16251 allocs/op

# after
BenchmarkRegistryListRepeated-32   5228-5832 iters 201-208 µs/op     ~188-191 KiB/op    1626 allocs/op
```

Representative improvement: ~10x faster repeated listing and ~93% fewer allocations.

## Verification

```text
go test ./internal/skills
go test ./internal/skills -run '^$' -bench BenchmarkRegistryListRepeated -benchmem -count=5
go build ./...
go test ./...
```

### DIFF
--- a/internal/skills/registry.go
+++ b/internal/skills/registry.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"path/filepath"
 	"sort"
+	"strconv"
 	"strings"
 )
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #616, opened 2026-05-11
fix: Server decodes full uploaded image payloads even when the original base64 could be reused

sam-saffron-jarvis:
## What

- preflight uploaded base64 payload sizes from their encoded length before allocating decoded bytes
- keep small LLM-native inline images in their original base64 form instead of eagerly decoding and re-encoding them
- add regression coverage for the fast path and the oversize-before-decode guard

## Why

The web server was fully decoding every uploaded image before enforcing the attachment size limit, and it also routed inline LLM-native images through that decode path even when the original base64 could be sent unchanged. In image-heavy sessions that creates avoidable CPU work and higher peak memory usage on the server.

### DIFF
--- a/cmd/serve_protocol.go
+++ b/cmd/serve_protocol.go
@@ -111,15 +111,46 @@ const (
 	maxAttachmentBytes = 20 << 20 // 20 MB per file (decoded)
 )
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #617, opened 2026-05-11
fix: Gemini edit requests build giant base64+JSON copies in memory

sam-saffron-jarvis:
## What

Stream Gemini image-edit request bodies instead of prebuilding a giant in-memory base64+JSON payload.

- keep Gemini edit parts as raw image bytes until upload time
- write request JSON directly to the HTTP body, base64-encoding inline images on the fly
- preserve the existing buffered path for debug logging and text-only requests
- add a regression test that verifies inline-image edits stream the request body and still send the expected payload

## Why

Gemini edits were loading image bytes, then creating a second full-size base64 string copy for every input image, and then marshaling the whole request into another large JSON buffer before upload started. Multi-image edits could therefore hit large memory spikes and noticeable startup delay. Streaming the request body removes the extra pre-upload copy in the normal edit path while keeping the change tightly scoped.

### DIFF
--- a/internal/image/gemini.go
+++ b/internal/image/gemini.go
@@ -63,7 +63,7 @@ func (p *GeminiProvider) Edit(ctx context.Context, req EditRequest) (*ImageResul
 		parts = append(parts, geminiPart{
 			InlineData: &geminiInlineData{
 				MimeType: mimeType,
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #618, opened 2026-05-11
fix: Web UI eagerly base64-encodes and retains every attachment before upload

sam-saffron-jarvis:
## What

- stop mutating pending Web UI attachments into persistent base64 `data:` URLs before send
- materialize attachment payloads only into transient request parts used for `/v1/responses`
- keep sent message attachments lightweight by preserving their existing preview/file references instead of storing the encoded payload on the session message
- rebuild prior attachment turns on demand when the UI has to reconstruct conversation input, with coverage for both normal sends and attachment read failures

## Why

The old flow eagerly read every attachment into base64, replaced the preview blob URL with that larger string, dropped the original file, and then kept the base64 payload on message/session objects even after the request was sent. That caused long pre-send stalls on multi-attachment turns, multiple in-memory copies of the same file, and persistent client-side memory bloat.

This change keeps the expensive base64 payload transient while preserving current behavior for previews and previous-response recovery.

### DIFF
--- a/internal/serveui/static/app-stream.js
+++ b/internal/serveui/static/app-stream.js
@@ -2686,19 +2686,45 @@ const readFileAsDataURL = (file, signal) => new Promise((resolve, reject) => {
 });
 
 const materializeAttachmentDataURL = async (att, signal) => {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #619, opened 2026-05-11
fix: filesDir serving duplicates generated media under a new random filename on every call

sam-saffron-jarvis:
## What

Use a deterministic content-derived filename when `ensureFileServeable` has to materialize an approved file into `filesDir`, and reuse the existing copy when that content is already present.

Also add a regression test covering repeated calls for the same generated file so they return the same path and leave only one copied file in `filesDir`.

## Why

The old code always copied approved files to `filesDir` as `serve-<random>-<basename>`, so the same generated media could be recopied on every tool-event emission or replay. That wasted disk writes and storage, produced a new URL every time, and defeated browser caching when reopening or redownloading generated media.

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -541,22 +541,42 @@ func (s *serveServer) ensureFileServeable(filePath string) (string, bool) {
 	}
 	defer src.Close()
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #620, opened 2026-05-11
fix: Image serving re-copies the full source file before checking whether the hashed cache target already exists

sam-saffron-jarvis:
## What

- hash the approved source image first so `ensureImageServeable` can check the deterministic cache path before creating a temp file or copying bytes into the output directory
- only rewind and materialize a temp copy when the hashed destination does not already exist
- verify with a regression test that a repeated call reuses the existing materialized copy without mutating the output directory

## Why

Repeated history fetches and tool result replays were still creating and deleting a temp file after reading the full source image, even when the stable hashed copy had already been materialized. That adds avoidable disk I/O and latency on a hot path. This change preserves the stable content-derived naming while skipping the extra output-directory write work for cache hits.


SamSaffron:

PR #621, opened 2026-05-11
perf: reduce smooth buffer drain copying

sam-saffron-jarvis:
## What

Reduce `SmoothBuffer.NextWords` backlog drain work in the terminal streaming path.

The old implementation converted the entire pending buffer to a string, converted that whole string to `[]rune`, extracted a few words, then rebuilt the buffer with the remaining text on every 60fps tick. This made draining large provider bursts roughly quadratic in the unread backlog.

This PR changes `SmoothBuffer` to keep a byte slice plus read offset, scan only the emitted prefix using UTF-8 decoding, and compact occasionally. Compaction drops oversized backing arrays when a large burst has mostly drained, so the offset-based buffer does not retain peak memory indefinitely.

It also adds regression coverage for:

- appending after a partial read,
- compaction preserving unread bytes and shrinking oversized backing storage,
- long-word chunking by rune for multi-byte text,
- backlog drain benchmark coverage.

## Why

`SmoothBuffer` sits in the visible terminal streaming path for ask/chat. Fast providers can enqueue large bursts, and the UI then releases a handful of words per frame. Avoiding full-buffer copies and full-buffer rune conversion per frame makes streaming repaint lighter and less latency-prone under backlog.

## Evidence

Added `BenchmarkSmoothBufferDrainBacklog` and ran before/after with `-benchtime=100ms -count=3` on linux/amd64 i9-14900K.

Representative results:

| Backlog | Before | After |
| --- | ---: | ---: |
| 4KB | ~1.48-1.62 ms/op, ~2.17 MB/op, 1090 allocs/op | ~28.7-29.9 µs/op, ~20.3 KB/op, 196 allocs/op |
| 16KB | ~21.0-21.9 ms/op, ~34.5 MB/op, 4040 allocs/op | ~108.7-110.2 µs/op, ~83.8 KB/op, 689 allocs/op |
| 64KB | ~315-329 ms/op, ~538 MB/op, ~15840 allocs/op | ~449-480 µs/op, ~338 KB/op, 2656 allocs/op |

## Verification

- `gofmt -w internal/ui/smooth_buffer.go internal/ui/smooth_buffer_test.go internal/ui/smooth_buffer_bench_test.go`
- `go test ./internal/ui`
- `go test ./internal/ui -run '^$' -bench '^BenchmarkSmoothBufferDrainBacklog$' -benchmem -benchtime=100ms -count=3`
- `go build ./...`
- `go test ./...`

### DIFF
--- a/internal/ui/smooth_buffer.go
+++ b/internal/ui/smooth_buffer.go
@@ -5,6 +5,7 @@ import (
 	"sync"
 	"time"
 	"unicode"
+	"unicode/utf8"
 
 	tea "charm.land/bubbletea/v2"
 )
@@ -16,6 +17,8 @@ const (
 	SmoothMinWordsPerFrame = 1
 	SmoothMaxWordsPerFrame = 5
 	SmoothMaxWordLength    = 12 // Chunk words longer than this
+
+	smoothBufferCompactThreshold = 4096
 )
 
 // SmoothTickMsg is sent to trigger the next frame of smooth text rendering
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #622, opened 2026-05-11
perf: avoid duplicate skill registry scan

sam-saffron-jarvis:
## What

- Build the skill registry `List()` cache fingerprint during the first discovery scan instead of doing a separate pre-scan when the list cache is empty.
- Reuse the already-statted `SKILL.md`/`skill.md` manifest path when loading skill metadata during registry discovery.
- Add a cold `Registry.List()` benchmark alongside the existing repeated-list benchmark.

## Why

Short-lived CLI commands and completions commonly hit the skill catalog with a fresh registry, so the previous list cache could not help the first call. That first call fingerprinted every skill path, then immediately scanned the same directories again, then `IsSkillDir`/`LoadFromDir` statted the same manifest again before parsing metadata.

This keeps cache invalidation behavior for subsequent calls while removing redundant filesystem work from the first-use startup path.

## Evidence

Synthetic catalog: 1000 `.skills/*/SKILL.md` entries, running `term-llm skills` under `strace -qq -f -c -e trace=%file`.

- Before: 3038 `newfstatat`, 1019 `openat`, 4061 total file syscalls, 0.008045s reported syscall time.
- After: 1038 `newfstatat`, 1015 `openat`, 2057 total file syscalls, 0.002716s reported syscall time.

Focused benchmark after the change (`go test ./internal/skills -bench 'BenchmarkRegistryList(Cold|Repeated)$' -benchmem -run '^$' -count=3`):

- `BenchmarkRegistryListCold`: ~1.97-2.03 ms/op for 200 skills.
- `BenchmarkRegistryListRepeated`: ~233-259 µs/op for 200 skills.

## Verification

- `gofmt -w internal/skills/parser.go internal/skills/registry.go internal/skills/registry_test.go`
- `go build ./...`
- `go test ./...`
- `go test ./internal/skills -bench 'BenchmarkRegistryList(Cold|Repeated)$' -benchmem -run '^$' -count=3`

### DIFF
--- a/internal/skills/parser.go
+++ b/internal/skills/parser.go
@@ -280,7 +280,10 @@ func extractExtrasFromNode(node *yaml.Node) (map[string]any, error) {
 // LoadFromDir loads a skill from a directory containing SKILL.md.
 // If loadBody is false, only metadata is loaded (for discovery).
 func LoadFromDir(dir string, source SkillSource, loadBody bool) (*Skill, error) {
-	// Try SKILL.md first, then skill.md (with warning)
+	// Try SKILL.md first, then skill.md (with warning). Preserve the historical
+	// behavior of falling back to lowercase only when the preferred file is known
+	// to be absent, so permission/symlink errors on SKILL.md are surfaced by the
+	// open below instead of being masked by skill.md.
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #623, opened 2026-05-12
fix: Glob traversal re-Lstats every ancestor path on every filesystem access---BUG---

sam-saffron-jarvis:
## What

- make `EventToolExecEnd` emission in `executeSingleToolCall` best-effort/non-blocking by switching those sends to `TrySend`
- apply the same non-blocking behavior to the panic-recovery path in `executeSingleToolCallSafe`
- add a regression test covering parallel tool execution with a full event buffer so completed tools still return results instead of stalling workers

## Why

Parallel tool workers were sending `EventToolExecEnd` with a blocking channel send before returning their tool result. When the event buffer filled because of a slow SSE/WebSocket/UI consumer, completed workers could block at the finish line and stop picking up more work, collapsing effective tool concurrency and stretching agent latency. These end events were already best-effort in practice because send errors were ignored, so making them explicitly non-blocking fixes the stall without changing tool-result delivery.

### DIFF
--- a/internal/llm/engine.go
+++ b/internal/llm/engine.go
@@ -1613,7 +1613,7 @@ func (e *Engine) executeSingleToolCallSafe(ctx context.Context, call ToolCall, s
 	defer func() {
 		if r := recover(); r != nil {
 			errMsg := fmt.Sprintf("Error: tool panicked: %v", r)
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #624, opened 2026-05-12
fix: Serve mode rewrites the full session message history after every run

sam-saffron-jarvis:
## What

- stop serve mode from doing an unconditional full `ReplaceMessages` rewrite after every successful run
- keep the existing incremental `AddMessage`/`UpdateMessage` persistence path as the normal success path
- only fall back to a final full snapshot when incremental persistence never started, missed writes, or left the in-flight assistant row dirty
- add regression coverage proving successful runs no longer rewrite the whole transcript and that initial-snapshot retry behavior still works

## Why

Serve mode was rebuilding and rewriting the entire session transcript after each run even though the same run had already been persisted incrementally during streaming. In long, tool-heavy sessions this turned every new turn into O(total history) serialization and database work while the per-session runtime lock was still held. Skipping the redundant rewrite keeps persistence incremental on the happy path, preserving correctness while preventing later turns from getting progressively slower.

### DIFF
--- a/cmd/serve_runtime.go
+++ b/cmd/serve_runtime.go
@@ -608,6 +608,7 @@ func (rt *serveRuntime) run(ctx context.Context, stateful bool, replaceHistory b
 	var produced []llm.Message
 	var producedMu sync.Mutex
 	var lastAppendedIdx int // tracks how many produced messages have been incrementally persisted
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #625, opened 2026-05-12
fix: Every shell/custom tool teardown scans the entire /proc tree

sam-saffron-jarvis:
## What

Avoid the host-wide `/proc` scan on every shell/custom tool teardown.

This change keeps the existing process-group kill, but only falls back to `killTaggedDescendants` when it is actually needed:
- the command was cancelled/timed out, or
- a lightweight `/proc/<pid>/task/<pid>/children` monitor observed a descendant escape the tool's process group while the command was running.

It also adds tests covering the fast path and the cancellation path.

## Why

`shell`, `custom_tool`, and `run_agent_script` all sit on a hot execution path. Before this fix, even tiny commands always paid for a full `/proc` walk plus reads of every process's `environ` during cleanup.

On busy hosts that adds avoidable latency proportional to total system process count rather than the work the tool actually did. The new behavior preserves the leaked-descendant cleanup fallback for real escape cases while removing that overhead from normal command completions.

### DIFF
--- a/internal/tools/exec_helpers.go
+++ b/internal/tools/exec_helpers.go
@@ -10,6 +10,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync/atomic"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing without integrating as-is. The optimization is attractive, but skipping the nonce /proc scan on successful commands can miss fast escaped descendants (for example double-fork or setsid races) and regress the tool cleanup guarantee. Keeping the conservative cleanup behavior for now.

PR #626, opened 2026-05-12
fix: make stream adapter cancellation responsive

sam-saffron-jarvis:
## What

Make `ui.StreamAdapter.ProcessStream` use context-aware sends for every translated stream event. If the downstream event channel is full and the caller cancels the stream context, the adapter now exits and closes the event channel instead of staying blocked on a raw channel send.

## Why

The adapter intentionally applies backpressure to provider/tool streams, but raw `a.events <- ...` sends meant cancellation could not unblock the adapter once the UI/JSON consumer stopped draining the channel. In real sessions this can keep provider streaming and tool execution cleanup goroutines alive longer than necessary after an interrupt/disconnect.

## Evidence

Added `TestStreamAdapterCancellationUnblocksBlockedSend`, which fills a one-slot adapter buffer, cancels the context while the second event send is blocked, and asserts `ProcessStream` returns promptly.

Before the fix:

```text
--- FAIL: TestStreamAdapterCancellationUnblocksBlockedSend (0.25s)
    stream_adapter_test.go:200: ProcessStream did not return after context cancellation while blocked on send
```

After the fix:

```text
go test ./internal/ui -run TestStreamAdapterCancellationUnblocksBlockedSend -count=5
ok  github.com/samsaffron/term-llm/internal/ui  0.016s
```

Also verified:

```text
go build ./...
go test ./...
```

### DIFF
--- a/internal/ui/stream_adapter.go
+++ b/internal/ui/stream_adapter.go
@@ -65,35 +65,57 @@ func (a *StreamAdapter) Stats() *SessionStats {
 //
 //	go adapter.ProcessStream(ctx, stream)
 func (a *StreamAdapter) ProcessStream(ctx context.Context, stream llm.Stream) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
 	defer close(a.events)
+
+	emit := func(event StreamEvent) bool {
+		if ctx.Err() != nil {
+			return false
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #627, opened 2026-05-12
perf: avoid redundant markdown normalization

sam-saffron-jarvis:
## What

- Avoid full input copies in the ANSI markdown renderer when tab normalization is configured but the input contains no tabs.
- Skip regex newline normalization and inline escape/entity decoding when the rendered/text fragments do not contain the relevant markers.
- Let the streaming markdown renderer pass its accumulated markdown buffer through without `bytes.ReplaceAll` when the stream has not seen tabs.
- Add focused benchmarks for ANSI markdown rendering and streaming markdown rendering in the common no-tab path.

## Why

Terminal rendering frequently re-renders markdown while responses stream. Most model output contains no tabs, no 3+ newline runs, and many inline text fragments without escapes/entities. The previous path still performed full-buffer copies or regex passes in those no-op cases.

## Evidence

On this machine (`Intel(R) Core(TM) i9-14900K`, `go test -benchmem -count=5`):

- `go test ./internal/render/markdown -run '^$' -bench BenchmarkANSIRenderLargeNoTabs -benchmem -count=5`
  - Before: ~2.42-2.49 ms/op, ~1.507-1.512 MB/op, 16075-16076 allocs/op
  - After: ~2.36-2.44 ms/op, ~1.276-1.279 MB/op, 12988-12989 allocs/op
- `go test ./internal/ui/streaming -run '^$' -bench BenchmarkStreamRendererLargeNoTabs -benchmem -count=5`
  - Before: ~82.1-84.3 ms/op, ~46.15-46.27 MB/op, 528829-528849 allocs/op
  - After: ~79.4-81.1 ms/op, ~44.01-44.17 MB/op, 428634-428659 allocs/op

## Verification

- `go test ./internal/render/markdown ./internal/ui/streaming`
- `go build ./...`
- `go test ./...`

### DIFF
--- a/internal/render/markdown/ansi.go
+++ b/internal/render/markdown/ansi.go
@@ -1,6 +1,7 @@
 package markdown
 
 import (
+	"bytes"
 	"fmt"
 	"html"
 	"regexp"
@@ -22,6 +23,8 @@ import (
 )
 
 // Renderer renders complete markdown content for a terminal target.
+// Implementations must treat source as read-only and must not retain it after
+// Render returns; streaming callers may pass slices backed by reusable buffers.
 type Renderer interface {
 	Render(source []byte) ([]byte, error)
 	Resize(width int)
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #631, opened 2026-05-13
fix: Hybrid memory search does a full embedding-table scan and in-memory rerank

sam-saffron-jarvis:
## What

- change `VectorSearch` to keep only the current top-k matches in memory instead of collecting and sorting every scored embedding row
- stop loading full fragment payloads during the initial embedding scan; only fetch fragment bodies/metadata for the final top-k fragment IDs
- preserve the existing provider/model/dimensions filtering and final score ordering semantics

## Why

Hybrid memory search always calls this path, so the old implementation got slower as the embedding table grew. It joined every matching embedding to full fragment content, unmarshaled every vector, stored all scored matches in Go, sorted the full result set, and only then truncated to the requested limit. Bounding the in-memory rerank and deferring full fragment loads to finalists cuts a lot of avoidable I/O and memory churn on large memory stores while keeping behavior targeted and low-risk.

### DIFF
--- a/internal/memory/store.go
+++ b/internal/memory/store.go
@@ -10,7 +10,6 @@ import (
 	"math"
 	"os"
 	"path/filepath"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


SamSaffron:

PR #632, opened 2026-05-13
perf: avoid double decoding response websocket events

sam-saffron-jarvis:
## What

- Add a fast path for extracting known top-level Responses WebSocket `type` fields without a preliminary full `json.Unmarshal`.
- Reuse the same event type extractor for SSE payloads that omit an explicit `event:` line.
- Fall back to `json.Unmarshal` for uncommon field orderings, unknown event types, duplicate-key compatibility cases, and malformed JSON.
- Add unit coverage for nested/escaped/malformed event-type extraction and a benchmark for WebSocket text-delta dispatch.

## Why

Responses-over-WebSocket previously decoded every frame once just to read the envelope `type`, then decoded the same payload again in the shared event handler. Text deltas are the hottest streaming event path, so removing that extra decode lowers per-frame CPU and allocations.

## Evidence

Benchmark on this machine (`go test ./internal/llm -run '^$' -bench BenchmarkResponsesWebSocketTextDeltaDispatch -benchmem`):

- Before: ~905-947 ns/op, 512 B/op, 12 allocs/op
- After: ~510-523 ns/op, 280 B/op, 7 allocs/op

## Verification

- `go test ./internal/llm -run 'TestResponses(JSONEventType|ClientStream_|WebSocket)' -count=1`
- `go test ./internal/llm -run '^$' -bench BenchmarkResponsesWebSocketTextDeltaDispatch -benchmem -count=5`
- `go test ./...`
- `go build ./...`

### DIFF
--- a/internal/llm/responses_event_handler.go
+++ b/internal/llm/responses_event_handler.go
@@ -40,14 +40,118 @@ func (h *responsesStreamEventHandler) OutputItems() []ResponsesInputItem {
 	return append([]ResponsesInputItem(nil), h.outputItems...)
 }
 
+// responsesJSONEventType reads the top-level Responses event type without
+// unmarshalling the full event envelope on the common WebSocket path where a
+// known "type" is the first object field. Frames still get decoded by
+// HandleJSONEvent below; this avoids an extra json.Unmarshal on hot events while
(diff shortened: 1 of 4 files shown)


SamSaffron:

PR #634, opened 2026-05-14
feat: use durable response ids for session continuations

sam-saffron-jarvis:
## What

Replaces the first-party `/v1/sessions/:id/messages` append endpoint with durable message-backed `/v1/responses` continuations.

- Adds durable `previous_response_id` tokens backed by persisted message rows (`resp_msg_<messages.id>`).
- Keeps live/random response run IDs for streaming resume/cancel.
- Makes the web UI post to `/v1/responses` again, with one new user message and `previous_response_id` when continuing.
- Removes the browser history replay path and the special append endpoint.
- Rejects stale durable continuations (`continue 100` when the session tail is already `101`) with `409` and leaves the DB unchanged.
- Rejects full-history/replay-shaped input with message-backed continuations before provider calls.
- Rejects conflicting `session_id` headers when the durable response id belongs to another session.
- Adds `messages.turn_index` for longer-term debugging/stats.

## Why

The web UI should be able to use the normal Responses surface without becoming a shadow session database. `previous_response_id` is now durable for persisted sessions because it points at the current persisted message tail, not an in-memory run mapping.

External fresh `/v1/responses` semantics remain intact: no `previous_response_id` still means fresh conversation/replace input.

## Verification

```text
go test ./...
go build ./...
node internal/serveui/static/app_stream_test.js
```

### DIFF
--- a/cmd/serve_ask_user_state_handler.go
+++ b/cmd/serve_ask_user_state_handler.go
@@ -96,6 +96,10 @@ func (s *serveServer) handleSessionState(w http.ResponseWriter, r *http.Request,
 		resp["reasoning_effort"] = persistedEffort
 	}
 
+	if lastResponseID := s.latestDurableResponseIDForSession(r.Context(), sessionID); lastResponseID != "" {
+		resp["lastResponseId"] = lastResponseID
+	}
+
 	if s.responseRuns != nil {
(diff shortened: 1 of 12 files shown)


SamSaffron:
Applying these changes directly on main locally to resolve conflicts with recent session stats/schema work.

PR #636, opened 2026-05-14
fix: Streamed prose can disappear until a newline/tool boundary because incomplete text only renders from the markdown stream buffer

sam-saffron-jarvis:
## What

- add a `PreviewUnflushed` path for streaming text segments that keeps already-rendered markdown blocks and appends a raw preview of the current incomplete prose block
- switch `RenderSegmentsWithLeadingAndImageRenderer` to use that preview instead of only `RenderedUnflushed()`
- add coverage for visible pending prose during streaming while still suppressing unstable table/list previews

## Why

`TextSegmentRenderer` uses the non-partial streaming markdown renderer, so incomplete paragraphs often remain buffered until a newline or tool boundary. The View path was only reading `RenderedUnflushed()`, which made streamed prose appear to freeze and then jump in large chunks. This change preserves smooth token-by-token prose updates without forcing snapshot refreshes in the hot render path or exposing unstable table/list partials.

### DIFF
--- a/internal/ui/segment.go
+++ b/internal/ui/segment.go
@@ -701,7 +701,7 @@ func RenderSegmentsWithLeadingAndImageRenderer(leading *Segment, segments []*Seg
 			if seg.Complete && seg.Rendered != "" {
 				rendered = seg.Rendered
 			} else if seg.StreamRenderer != nil {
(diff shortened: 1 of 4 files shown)


SamSaffron:

PR #637, opened 2026-05-14
fix: In-progress assistant text is dropped from the render path until a newline or boundary completes it

sam-saffron-jarvis:
## What

- switch the live `SegmentText` render path to use a view-specific streaming preview instead of only `RenderedUnflushed()`
- teach the streaming renderer to build a safe full-document preview for the current incomplete block without changing scrollback flush behavior
- keep pending list/table prefixes on the old hidden path, and add regression coverage for in-progress paragraph rendering and committed-to-pending block spacing

## Why

The TUI was dropping ordinary in-progress assistant prose from the render path until a newline, tool boundary, or final flush committed the block. That made streaming responses look frozen and then jump in chunks. This change keeps normal prose visible immediately while preserving the existing committed-only flush semantics and avoiding noisy previews for incomplete list/table structures.


SamSaffron:

PR #638, opened 2026-05-14
fix: Live streamed text is buffered until newline/tool boundary instead of appearing token-by-token

sam-saffron-jarvis:
## What

- enable partial streaming rendering for `TextSegmentRenderer` so in-progress prose appears in the live TUI before a newline or tool boundary
- teach the streaming renderer to support flowing, resettable snapshot updates for partial text while keeping list-marker/table previews suppressed
- keep scrollback flushes committed-block-only by tracking a committed rendered snapshot separately from the live partial preview
- add/update tests covering visible partial prose, markdown completion, list/table suppression, and scrollback safety

## Why

The live chat view was reading `RenderedUnflushed()` from a renderer created without partial rendering, so ordinary token deltas inside the current line often produced no visible output until a newline, tool boundary, or final flush. That made assistant text appear frozen and then jump in large chunks. This change restores token-by-token perceived streaming without regressing scrollback flushing or incomplete list/table handling.

### DIFF
--- a/internal/ui/segment_flush_test.go
+++ b/internal/ui/segment_flush_test.go
@@ -393,21 +393,21 @@ func TestPartialTextFlushSpacing(t *testing.T) {
 	}
 }
 
-func TestRenderUnflushed_HidesPendingParagraphContent(t *testing.T) {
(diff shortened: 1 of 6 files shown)


SamSaffron:

PR #639, opened 2026-05-14
fix: Pending-tool wave animation forces a full alt-screen viewport content reset every 50ms

sam-saffron-jarvis:
## What

- stop treating pending-tool wave animation ticks as viewport content changes in alt-screen mode
- cache split history lines and synthesize wave-only frames directly for the visible viewport instead of calling `viewport.SetContent(...)`
- add regression coverage that wave ticks still repaint the animated row without advancing `contentVersion`/`lastRenderedVersion`

## Why

Pending-tool wave animation was bumping `contentVersion` on every 50ms tick, which forced a full alt-screen viewport content reset even though only the existing tool row styling changed. On long sessions that means repeatedly rebuilding and re-sending the whole history + stream tail through Bubble Tea, causing avoidable tool-execution jank and rougher scrolling. This keeps the animation smooth while leaving the expensive viewport content update path for real content changes only.

### DIFF
--- a/internal/tui/chat/chat.go
+++ b/internal/tui/chat/chat.go
@@ -206,16 +206,17 @@ type Model struct {
 
 	// Render cache for alt screen mode (avoids re-rendering unchanged content)
 	viewCache struct {
-		historyContent      string // Cached rendered history
-		historyMsgCount     int    // Number of messages when cache was built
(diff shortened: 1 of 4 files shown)


SamSaffron:

PR #640, opened 2026-05-14
fix: Subagent progress uses one goroutine per UI event, removing backpressure and allowing message storms

sam-saffron-jarvis:
## What

- stop wrapping each subagent progress `Program.Send` call in its own goroutine
- send `spawn_agent` progress updates synchronously in both the chat TUI and ask renderer paths
- keep the fix narrowly scoped to the subagent progress callback wiring

## Why

`spawn_runner` emits `SubagentEventText` for every streamed text delta. Launching a fresh goroutine for every UI event removes Bubble Tea's natural backpressure, so a busy subagent can build up an unbounded backlog of goroutines and stale progress messages whenever the UI falls behind.

Sending these progress messages synchronously restores backpressure, preventing message storms from amplifying latency and jitter in the parent UI. The same callback pattern existed in the ask renderer, so it was fixed there as well to avoid the same behavior outside chat.


SamSaffron:

PR #641, opened 2026-05-14
perf: reduce memory vector search row loading

sam-saffron-jarvis:
## What

Optimize memory vector search so it no longer loads full fragment rows for every matching embedding before ranking.

VectorSearch now:
- scans only fragment id, updated_at, and embedding vector while scoring;
- keeps a bounded top-k heap instead of retaining and sorting every match;
- fetches heavyweight fragment fields (including content) only for the winning candidates;
- batches the follow-up fragment fetch so large limits do not exceed SQLite bind-variable limits.

A regression test covers a limit larger than the fragment fetch batch size, and a large-corpus benchmark captures the intended memory-search workload.

## Why

Hybrid memory search usually asks for a small candidate set (24), but the previous VectorSearch path selected large content bodies and built ScoredFragment values for every matching embedding before slicing to the limit. As memory grows, that creates avoidable row I/O, allocations, and sorting work on the interactive memory search path.

## Evidence

Representative benchmark: 3,000 fragments, 128-dimensional embeddings, ~2.8KB fragment content, limit 24.

Before:

```
BenchmarkStoreVectorSearchLargeCorpus-32  14-15  74.5-77.1 ms/op  45.6 MB/op  117044-117045 allocs/op
```

After:

```
BenchmarkStoreVectorSearchLargeCorpus-32  15-16  66.4-68.0 ms/op  23.6 MB/op  63739-63744 allocs/op
```

That is roughly 10% lower latency, 48% less allocated memory, and 45% fewer allocations for the benchmarked path.

## Verification

- `go test ./internal/memory -run 'TestStoreVectorSearch'`
- `go test ./internal/memory -run '^$' -bench BenchmarkStoreVectorSearchLargeCorpus -benchmem -count=5`
- `go test ./...`
- `go build ./...`

### DIFF
--- a/internal/memory/store.go
+++ b/internal/memory/store.go
@@ -1,6 +1,7 @@
 package memory
 
 import (
+	"container/heap"
 	"context"
 	"crypto/rand"
 	"database/sql"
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #645, opened 2026-05-15
fix: open search-only sidebar sessions

sam-saffron-jarvis:
## What

Fix sidebar search results that exist only in `sidebarSearchResults` but not yet in local `state.sessions`.

When opening one of those results, copy the search result into local session state before continuing with normal server-only hydration.

## Why

Recent sessions worked because they were already present in local sidebar state. Older search results came back from `/v1/sessions/search` but `switchToSession()` immediately bailed out if the ID was missing from `state.sessions`, so the click looked dead.

## Test

- `node internal/serveui/static/app_sessions_test.js`
- `go test ./...`


SamSaffron:

PR #646, opened 2026-05-15
fix: use exponential retry backoff

sam-saffron-jarvis:
## What

- Change provider retry backoff from linear+jitter to exponential+jitter.
- Preserve `Retry-After` numeric handling and add HTTP-date parsing for wrapped retry header text.
- Strengthen Responses WebSocket fallback coverage: after WebSocket setup exhausts its retry budget and falls back to HTTP, assert subsequent streams stay on HTTP instead of probing WebSocket again.

## Why

The comparative pass showed two things:

1. I was wrong about term-llm lacking WS→HTTP fallback — it already exists in `ResponsesClient.Stream`.
2. The generic provider retry loop was still linear, while the other agents mostly use exponential backoff for transient failures.

This makes the generic retry policy less hammer-y under outages, handles more real `Retry-After` formats, and locks in the existing WebSocket fallback behavior with a regression test.

## Validation

- `go test ./internal/llm`
- `go test ./...`
- `go build ./...`

### DIFF
--- a/internal/llm/responses_websocket_test.go
+++ b/internal/llm/responses_websocket_test.go
@@ -363,12 +363,14 @@ func TestResponsesClientWebSocketConnectFailureFallsBackToHTTP(t *testing.T) {
 	defer func() { responsesWebSocketBaseBackoff = oldBackoff }()
 
 	var wsAttempts atomic.Int32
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #648, opened 2026-05-15
fix: Skills startup scans the entire skill catalog twice before first prompt

sam-saffron-jarvis:
## What

- replace the eager `HasAnySkill()` startup probe in `internal/skills/setup.go` with a single `Registry.List()` call
- keep returning `nil` when no valid skills are available, but retain the discovered skill list on `Setup`
- reuse that preloaded catalog in `EnsurePromptMetadata()` so first-prompt skill XML generation does not rescan and reparse the skill trees

## Why

Skills are enabled by default, so startup was walking the configured skill paths twice before the first prompt: once to answer "are there any skills?" and again to build `<available_skills>`. Reusing the initial catalog keeps the fix tightly scoped while removing duplicate directory reads and manifest parsing from cold startup.

### DIFF
--- a/internal/skills/setup.go
+++ b/internal/skills/setup.go
@@ -21,6 +21,7 @@ type Setup struct {
 	alwaysEnabled        []string
 	metadataBudgetTokens int
 	maxVisibleSkills     int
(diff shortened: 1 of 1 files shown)


SamSaffron:

PR #649, opened 2026-05-15
fix: MCP tool completion cold-starts external servers on TAB

sam-saffron-jarvis:
## What

Stop `term-llm mcp run` shell completion from starting MCP servers on a cache miss. `MCPRunArgCompletion` now returns cached tool completions only, and otherwise exits immediately without spawning external processes. Added tests to verify completion stays on the cache-only path and still returns cached tool parameter completions.

## Why

Shell completion is on the interactive cold path, so a first `<TAB>` must stay instant. The previous behavior could block for up to 10 seconds while starting an MCP server and handshaking just to compute completions, which made tab completion feel hung and launched external processes unexpectedly.

### DIFF
--- a/cmd/completion.go
+++ b/cmd/completion.go
@@ -1,10 +1,8 @@
 package cmd
 
 import (
-	"context"
 	"sort"
 	"strings"
-	"time"
 
 	"github.com/samsaffron/term-llm/internal/config"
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #650, opened 2026-05-15
fix: Built-in agents always extract embedded resources to disk during prompt setup

sam-saffron-jarvis:
## What

- skip builtin resource extraction during prompt setup unless the builtin prompt actually needs extracted files
- continue extracting when the prompt uses `{{resource_dir}}` or builtin-relative `{{file:...}}` includes
- add regression coverage for the skipped-extraction path and both extraction-required cases

## Why

`agentPromptTemplateContextAndBaseDir` was extracting embedded builtin resources on every launch, even for common builtins like `@developer`, `@reviewer`, `@editor`, and `@commit-message` that do not use extracted files. That caused avoidable filesystem work before the first visible response. This change removes that startup churn while preserving the prompts that genuinely depend on extracted resources.

### DIFF
--- a/cmd/session.go
+++ b/cmd/session.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #651, opened 2026-05-15
perf: stream read_file long lines in chunks

sam-saffron-jarvis:
## What

Stream `read_file` line-numbered output with `bufio.Reader.ReadSlice` fragments instead of `ReadString`, so a single oversized line can be capped without allocating or scanning the whole logical line.

This keeps the existing line-numbered formatting but emits each line number once, appends line content in chunks, and stops immediately when the configured byte/line limit is reached.

## Why

`ReadString('\n')` must accumulate an entire logical line before the existing output byte cap can apply. For minified/generated files with multi-MB single lines, `read_file` paid the allocation and latency cost even when only the first 50KB (or less) could be returned.

## Evidence

Focused benchmark on an 8 MiB single-line file with `MaxBytes=1024`:

- Before: `2.17–3.33 ms/op`, `16,943,686–16,944,245 B/op`, `2098–2099 allocs/op`
- After: `9.67–10.50 µs/op`, `8,816 B/op`, `37 allocs/op`

Benchmark command:

```bash
go test ./internal/tools -run '^$' -bench BenchmarkReadFileToolHugeSingleLineCapped -benchmem -benchtime=10x -count=3
```

## Tests

- Added regression coverage for stopping within the first chunk after a byte cap.
- Added regression coverage for reassembling untruncated lines split across the bufio buffer.
- `go test ./internal/tools -run 'Test(ReadFileTool|StreamLineNumberedRange)' -count=1`
- `go build ./...`
- `go test ./...`

### DIFF
--- a/internal/tools/read.go
+++ b/internal/tools/read.go
@@ -172,8 +172,10 @@ func streamLineNumberedRange(ctx context.Context, reader *bufio.Reader, startLin
 	writtenLines := 0
 	truncated := false
 	byteCapped := false
-	emittedAny := false
+	sawData := false
 	lastEndedNewline := false
+	lineOpen := false
+	lineWriting := false
 
 	appendOutput := func(s string) {
 		if byteCapped {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #652, opened 2026-05-15
perf: speed up global jobs run summaries

sam-saffron-jarvis:
## What

Add a global covering index for Jobs V2 run-summary listing:

- keeps the existing job-scoped covering index for `/v2/runs?job_id=...`
- adds `idx_job_runs_v2_summary_created` for all-job summaries ordered by `created_at DESC`
- adds an EXPLAIN test to verify the global summary query uses the covering index and avoids a temp B-tree sort
- adds a benchmark for global run-summary listing across 100 jobs / 100k runs

## Why

The CLI/status path fetches recent run summaries across all jobs (`/v2/runs?summary=true` without `job_id`). The existing covering index starts with `job_id`, so SQLite could not use it to satisfy the global `ORDER BY created_at DESC`; it scanned/sorted the summary rows via a temp B-tree.

This keeps the endpoint behavior unchanged while reducing latency for real jobs/status listing as run history grows.

## Evidence

Representative benchmark over 100 jobs / 100k runs, listing the newest 100 global summaries:

- Before: ~10.6-11.9 ms/op, ~227 KB/op, 5246 allocs/op
- After: ~1.55 ms/op, ~229 KB/op, 5246 allocs/op

The new EXPLAIN test asserts the global summary query uses `USING COVERING INDEX idx_job_runs_v2_summary_created` and does not emit `USE TEMP B-TREE`.

## Verification

- `go test ./cmd -run 'TestJobsV2(RunSummary|GlobalRunSummary)'`
- `go test ./cmd -run '^$' -bench '^BenchmarkJobsV2ListRunSummariesGlobal$' -benchmem -count=3`
- `go build ./...`
- `go test ./...`

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -613,13 +613,23 @@ func newJobsV2Manager(dbPath string, workers int, llmExec serveJobsExecutor) (*j
 	for _, migration := range migrations {
 		_, _ = db.Exec(migration)
 	}
-	// Replace the narrow job_id/created_at index with a covering index for the
+	// Replace the narrow job_id/created_at index with covering indexes for the
 	// run-summary API. Large stdout/stderr/thinking/response columns sit before
-	// summary metadata in the table record; without a covering index, SQLite
-	// walks overflow payload pages just to list recent runs.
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #653, opened 2026-05-16
fix: Every follow-up send blocks on a full server resync before the POST stream starts

sam-saffron-jarvis:
## What

- stop `sendMessage` from waiting on a full session message-history reload before posting a normal follow-up
- keep the preflight continuation refresh, but call `syncActiveSessionFromServer(..., { skipMessagesFetch: true })` so it only refreshes lightweight runtime state like `lastResponseId` / active-run status
- extend the app-stream regression test to assert the preflight sync stays state-only

## Why

Follow-up sends in established chats were paying an unnecessary latency tax before the POST stream could even start. The preflight sync was reloading the session's full paginated message history whenever the server considered the run idle, which made first-token latency noticeably worse on longer chats. This change preserves the stale-continuation guard without forcing that expensive history refetch on every send.

### DIFF
--- a/internal/serveui/static/app-stream.js
+++ b/internal/serveui/static/app-stream.js
@@ -3264,7 +3264,10 @@ const sendMessage = async (options = {}) => {
   );
   if (shouldRefreshContinuation && typeof app.syncActiveSessionFromServer === 'function') {
     try {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #654, opened 2026-05-16
fix: Session history loading fetches every page serially before rendering anything

sam-saffron-jarvis:
## What

- stop blocking session switches on a full serial history walk for server-only sessions
- render the first fetched page as soon as it arrives while the rest of the history continues loading
- prefetch remaining history pages in small parallel batches instead of one-by-one

## Why

Large chat histories were paying one network round trip per 200-message page before showing anything. That made opening long chats, hydrating server-only sessions, and post-run reconciliation get slower as histories grew. Rendering the initial page early improves perceived responsiveness, and batching later page fetches cuts total wait time without a broader refactor.


SamSaffron:

PR #655, opened 2026-05-16
fix: Sidebar/session listing does a per-session COUNT(*) subquery on every List call

sam-saffron-jarvis:
## What

Persist visible `message_count` on the `sessions` row and switch `List()` to read it directly instead of running a correlated `COUNT(*)` subquery per returned session.

This adds a schema migration that backfills counts for existing databases, installs SQLite triggers to keep the count in sync for message inserts/deletes/role changes, and keeps the old subquery path only as a compatibility fallback for read-only legacy schemas.

## Why

The web sidebar and session status polling hit `List()` frequently. Computing `message_count` with a per-session subquery turns those cheap refreshes into repeated scans over `messages`, which slows sidebar load and update paths as session history grows.

Persisting the visible count shifts that work to the write path, making repeated sidebar/session listing calls stay O(1) per session row while preserving the existing `user`/`assistant` count semantics.


SamSaffron:

PR #656, opened 2026-05-16
fix: Non-append transcript updates rebuild the entire chat DOM and reparse all markdown

sam-saffron-jarvis:
## What

- keep the append fast path for pure tail growth, but reconcile same-session non-append transcript changes by message id instead of clearing `elements.messages`
- reuse existing message nodes when their render-relevant data is unchanged, only update/create/remove the specific nodes that actually changed
- track per-message render keys so unchanged assistant messages keep their existing markdown DOM instead of being reparsed
- add regression coverage for history insertion reuse and same-length assistant-content refreshes

## Why

Non-append transcript updates currently tear down and rebuild the entire chat subtree for the active session. On large conversations that reparses every assistant markdown block, recreates the whole DOM, and causes visible jank during history merges, reconnect recovery, and hydration. Reusing keyed message nodes makes those updates much cheaper while preserving the existing full rebuild behavior on actual session switches.

### DIFF
--- a/internal/serveui/static/app-render.js
+++ b/internal/serveui/static/app-render.js
@@ -1863,6 +1863,93 @@ const updateToolGroupNode = (message) => {
 
 let _lastRenderedSessionId = null;
 let _lastRenderedMessageIds = [];
+let _lastRenderedMessageKeys = new Map();
+
+const messageRenderKey = (message) => {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #657, opened 2026-05-16
fix: Every follow-up message blocks on a full server sync before streaming can begin

sam-saffron-jarvis:
## What

Stop blocking normal follow-up sends on an unconditional preflight session sync.

This narrows the pre-send refresh to the cases that actually need it: when a session has prior context but is missing a local `lastResponseId`. That refresh now also skips message-history reloads.

For sessions that already have a local continuation cursor, `sendMessage` posts immediately. If the server later reports that the cursor is stale or that the session is already busy, the UI does a lightweight state refresh and retries once with the existing drafted user message instead of surfacing an avoidable error.

Also add focused browser-side tests covering the no-preflight fast path and stale-continuation recovery.

## Why

The previous flow awaited `syncActiveSessionFromServer(session, false)` for almost every existing conversation before starting `/v1/responses`. That meant first-token latency for a simple follow-up was gated on an extra `/state` round trip and, when idle syncs reloaded history, potentially many paginated message fetches.

This change keeps the important correctness safeguards for missing/stale continuation IDs and hidden active runs, but removes the unconditional sync from the common case so follow-up prompts can start streaming sooner.

### DIFF
--- a/internal/serveui/static/app-stream.js
+++ b/internal/serveui/static/app-stream.js
@@ -52,6 +52,27 @@ const normalizeError = async (response) => {
   return { status: response.status, message };
 };
 
+const hasSessionContinuationContext = (session) => Boolean(
+  session && (
+    Number(session.number || 0) > 0
+    || (Array.isArray(session.messages) && session.messages.length > 0)
+  )
+);
+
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #658, opened 2026-05-16
perf: stop runaway shell output early

sam-saffron-jarvis:
## What

Stop runaway shell commands that exceed a raw stdout/stderr drain budget instead of continuing to stream discarded bytes until the timeout fires.

- Shares a hard raw-output budget across stdout and stderr for `ShellTool`.
- Cancels the command context once the budget is crossed, while still retaining the existing bounded captured output.
- Reports a clear `[Command stopped after exceeding output limit]` marker separately from real timeouts.
- Adds a regression test and benchmark for infinite-output shell commands.

## Why

The existing `limitedBuffer` protected memory by retaining only `MaxBytes`, but it returned success for every subsequent write. A command like `yes x` with a small output cap still ran until `timeout_seconds`, burning wall-clock time and CPU even though no more useful output would be sent to the model.

Representative benchmark with `yes x`, `MaxBytes=16`, `CumulativeHard=256`, `timeout_seconds=1`:

- Before: `BenchmarkShellToolRunawayOutputSmallLimit-32  3  1000577260 ns/op`
- After: `BenchmarkShellToolRunawayOutputSmallLimit-32  3  3327777 ns/op`

## Verification

- `go test ./internal/tools -run 'TestShellTool_RunawayOutputStopsAtHardLimit' -count=1`
- `go test ./internal/tools -run '^$' -bench BenchmarkShellToolRunawayOutputSmallLimit -benchtime=3x -count=1`
- `go build ./...`
- `go test ./...`

### DIFF
--- a/internal/tools/exec_helpers.go
+++ b/internal/tools/exec_helpers.go
@@ -10,6 +10,8 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"syscall"
 	"time"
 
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #659, opened 2026-05-16
build: exclude UI JS fixtures from embed

sam-saffron-jarvis:
## What

- Replace the broad `//go:embed static/*` serve UI embed with a production-asset allowlist plus `static/vendor`.
- Add a regression test that embedded static assets contain every non-test production file, but no `*_test.js` JS fixtures.

## Why

The previous embed glob shipped the browser JS test fixtures inside the production binary and included them in `AssetVersion()` hashing, so test-only edits could churn UI cache keys and the binary carried avoidable bytes.

## Evidence

- `internal/serveui/static/*_test.js`: 266,687 source bytes no longer embedded.
- Local `go build -o` binary size: 85,652,576 bytes before vs 85,386,840 bytes after (265,736 bytes smaller).
- Verified with `go test -count=1 ./internal/serveui`, `go build ./...`, and `go test ./...`.

### DIFF
--- a/internal/serveui/embed.go
+++ b/internal/serveui/embed.go
@@ -13,7 +13,12 @@ import (
 	"sync"
 )
 
-//go:embed static/*
+//go:embed static/index.html static/manifest.webmanifest static/icon-512.png static/sw.js
+//go:embed static/app.css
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #665, opened 2026-05-17
fix: ChatGPT eagerly rebuilds full Responses input on every turn, defeating lazy continuation

sam-saffron-jarvis:
## What

- stop the ChatGPT provider from eagerly materializing full Responses `input` on every turn
- pass raw `Messages` into `ResponsesClient` and mark that system messages are being sent separately via `instructions`
- teach lazy full-input reconstruction to omit system messages when `instructions` came from the transcript
- add a websocket continuation test covering instruction extraction plus previous-response reuse

## Why

ChatGPT was building the full Responses transcript up front before the client could decide whether only a continuation suffix was needed. On long follow-up turns, especially after tool use, that forced avoidable transcript scanning/allocation into the pre-request path and increased time-to-first-token. This keeps the cheaper message-based continuation path available while preserving the existing ChatGPT wire format when a full-history rebuild is actually required.

### DIFF
--- a/internal/llm/chatgpt.go
+++ b/internal/llm/chatgpt.go
@@ -235,19 +235,22 @@ func (p *ChatGPTProvider) Stream(ctx context.Context, req Request) (Stream, erro
 		tools = append([]any{ResponsesWebSearchTool{Type: "web_search"}}, tools...)
 	}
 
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #666, opened 2026-05-17
fix: Request-shape comparison does a full JSON marshal/unmarshal of tool config on every WebSocket turn

sam-saffron-jarvis:
## What

Replace the WebSocket request-shape comparison hot path with a direct JSON-like comparator for `Tools` and `ToolChoice` instead of round-tripping them through `json.Marshal`/`json.Unmarshal` on every streamed turn.

Also add tests covering:
- JSON-shape equality across struct-vs-map tool definitions
- order-insensitive comparison for string arrays inside tool schemas
- no `MarshalJSON` call during comparison

## Why

`prepareWebSocketContinuationLocked` runs on every streamed WebSocket turn to decide whether server-side response state can be reused. Large MCP tool schemas were being fully marshaled and unmarshaled before every request even when unchanged, adding avoidable CPU, allocations, and pre-stream latency in a hot path.

The new comparer preserves the previous JSON-like comparison behavior needed for tool config checks while avoiding the expensive encode/decode round-trip.

### DIFF
--- a/internal/llm/responses_websocket.go
+++ b/internal/llm/responses_websocket.go
@@ -9,7 +9,6 @@ import (
 	"net/http"
 	"net/url"
 	"reflect"
-	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #667, opened 2026-05-17
fix: WebSocket continuation rebuilds and deep-compares the entire transcript before every reused request

sam-saffron-jarvis:
## What

- stop rebuilding the full transcript in WebSocket continuation preparation when an incremental continuation is already available
- pass the continuation builder into the WebSocket preparation path and prefer it over local full-history prefix reconstruction
- preserve `previous_response_id` for WebSocket requests even when there is no local `wsLastRequest` baseline yet
- remove the now-unused WebSocket response-item baseline bookkeeping and add regression coverage for both the no-full-rebuild path and the no-local-baseline path

## Why

Responses-over-WebSocket already has server-side conversation state, so rebuilding and deep-comparing the entire local transcript before each follow-up adds avoidable pre-network latency that grows with chat size. Using the already-incremental continuation keeps follow-up requests cheap, improves time-to-first-token in long chats, and still falls back to full input when request shape changes or incremental input is unavailable.

### DIFF
--- a/internal/llm/responses_api.go
+++ b/internal/llm/responses_api.go
@@ -52,7 +52,6 @@ type ResponsesClient struct {
 	wsMu                    sync.Mutex
 	wsConn                  *websocket.Conn
 	wsLastRequest           *ResponsesRequest
-	wsLastResponseItems     []ResponsesInputItem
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #668, opened 2026-05-17
fix: Context trimming is O(n²) because it repeatedly rescans token counts while dropping messages

sam-saffron-jarvis:
## What

- stop `EstimateMessageTokens` from allocating and re-filtering by skipping event rows inline
- cache the preserved-prefix and remaining-conversation token totals in `doTrim`
- subtract the dropped message's token estimate as trimming advances instead of rescanning the whole suffix each iteration
- add coverage that token estimation still ignores event messages

## Why

`doTrim` previously called `EstimateMessageTokens(prefix)` and `EstimateMessageTokens(conv)` on every loop iteration while dropping old messages. Near the context limit that turned trimming into quadratic work over the longest conversations, adding avoidable latency right before provider requests start. Keeping running totals preserves the existing behavior while making trimming linear in the number of messages.

### DIFF
--- a/internal/llm/compaction.go
+++ b/internal/llm/compaction.go
@@ -54,23 +54,28 @@ func EstimateTokens(text string) int {
 // EstimateMessageTokens returns an approximate token count for a slice of
 // messages by summing all text content across parts.
 func EstimateMessageTokens(msgs []Message) int {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #669, opened 2026-05-17
fix: Responses tool schemas are reparsed and fully normalized on every request instead of being cached

sam-saffron-jarvis:
## What

Cache the lowered OpenAI Responses `parameters` map per immutable tool schema identity and `strict` flag instead of reparsing and renormalizing it on every request.

Also add a unit test covering cache reuse for the same schema in strict and non-strict modes.

## Why

Large MCP toolsets were paying the full `ParseToolJSONSchemaMap` + `ToOpenAIParameters` normalization cost on every turn even though `ToolSpec.Schema` is documented as immutable and the lowered output is deterministic for a given `strict` value. Reusing the lowered schema removes that repeated CPU/allocation work from request construction and improves TTFT for Responses requests with many or large tool schemas.

### DIFF
--- a/internal/llm/json_schema_lowering.go
+++ b/internal/llm/json_schema_lowering.go
@@ -1,5 +1,10 @@
 package llm
 
+import (
+	"reflect"
+	"sync"
+)
+
 // ToOpenResponsesParameters serializes the typed schema into the provider-neutral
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #670, opened 2026-05-17
speed up memory vector search

sam-saffron-jarvis:
## What

- Store newly generated memory embeddings as compact little-endian float64 blobs instead of JSON arrays.
- Decode both the new binary payloads and existing legacy JSON payloads, preserving compatibility with current memory DBs.
- Score binary vectors directly from SQLite row payloads during vector search and use `sql.RawBytes` to avoid an extra per-row blob copy.
- Update vector-search benchmarks to seed the current binary format and add regression coverage for legacy JSON decoding paths.

## Why

A whole-stack benchmark/profile pass found memory vector search spending most CPU in `encoding/json` float parsing/validation and allocating heavily while scanning embedding rows. New embeddings should avoid JSON parse overhead on every search.

## Evidence

Baseline profile/benchmark before this change:

```text
BenchmarkStoreVectorSearchLargeCorpus-32    15    68104829 ns/op    23586128 B/op    63742 allocs/op
```

CPU profile showed `encoding/json.checkValid`, `encoding/json.(*decodeState).literalStore`, and `internal/strconv.ParseFloat/readFloat` as the dominant samples.

After this change:

```text
BenchmarkStoreVectorSearchLargeCorpus-32              1046     5697180 ns/op    4458357 B/op    30857 allocs/op
BenchmarkStoreVectorSearchProviderModelFilter-32     14804      404316 ns/op     104990 B/op     3111 allocs/op
```

The provider/model filter benchmark baseline from the broad pass was `563569 ns/op`, `152773 B/op`, `4601 allocs/op`.

## Verification

```text
go test ./internal/memory
go test ./internal/memory -run '^$' -bench 'BenchmarkStoreVectorSearch(LargeCorpus|ProviderModelFilter)' -benchmem -benchtime=5s -count=1
go test ./...
go build ./...
```

### DIFF
--- a/internal/memory/store.go
+++ b/internal/memory/store.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/rand"
 	"database/sql"
+	"encoding/binary"
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #677, opened 2026-05-18
fix: Uploaded images are decoded, saved, optionally resized, re-encoded, and retained inline in one hot path

sam-saffron-jarvis:
## What

- stream base64 image uploads straight to the uploads directory instead of decoding the whole image into an intermediate raw buffer first
- reuse the saved original file for the rare >1 MB resize path, so small images avoid the extra decode/write hot-path work entirely
- add tests covering streamed upload saves and oversized payload rejection before decode

## Why

Image uploads were doing too much synchronous work in `parseUserMessageContent`: decode full base64 into memory, write another copy to disk, and only then decide whether resize work was needed. That inflated latency and memory usage on the request path, especially for larger or multiple image uploads. This change keeps the behavior the same while removing an unnecessary full decoded copy for the common case and coalescing the expensive path to only the uploads that actually need resizing.


SamSaffron:

PR #678, opened 2026-05-18
fix: Web image uploads are decoded, rewritten, resized, re-encoded, and kept inline all in one request

sam-saffron-jarvis:
## What

Change web image upload parsing to stop retaining inline base64 payloads in memory after request parsing. Uploaded images are still saved to disk, but the model-facing bytes are now staged via `InlineImagePath` and encoded lazily when a provider actually builds the request. For oversized images, the resized/compressed inline payload is written to a separate file so `ImagePath` can keep pointing at the original upload.

This also teaches the LLM provider adapters to load image bytes from `InlineImagePath`/`ImagePath` when `ImageData.Base64` is empty, and adds tests covering the new lazy-loading path.

## Why

The old flow kept multiple copies of each upload alive at once: the original data URL/base64 string, decoded raw bytes, optional resized bytes, and a freshly re-encoded inline base64 payload stored on the message part. Large or multi-image uploads therefore caused avoidable upload-time latency and memory spikes before the first model call started.

Deferring base64 encoding until provider request construction keeps request parsing much lighter while preserving the existing behavior of saving original uploads for tools and resizing oversized inline payloads for provider compatibility.

### DIFF
--- a/cmd/serve_protocol.go
+++ b/cmd/serve_protocol.go
@@ -222,13 +222,14 @@ func abbreviatePath(path string) string {
 // parseUserMessageContent builds a user llm.Message from a content field
 // that may be a plain string or an array of content parts (input_text, input_image, input_file).
 // Chat Completions-style text/image_url parts are also accepted.
-// Supported image types are sent inline to the LLM and also saved to disk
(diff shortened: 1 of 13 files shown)


SamSaffron:

PR #679, opened 2026-05-18
fix: Session storage writes full base64 image blobs into every message row

sam-saffron-jarvis:
## What

- stop session persistence from storing redundant base64 image blobs when the same image is already recoverable from a saved file path
- strip persisted base64 for user image parts with `ImagePath` and for tool-result image content parts with `ImagePath`
- teach provider image helpers to lazily reload image bytes from disk when a resumed session only has the file path
- have `view_image` include `ImagePath` in structured image content so its persisted tool results can be de-blobbed too
- add regression coverage for slim storage, pathless-image preservation, and image-path fallback loading

## Why

Session writes were serializing full image base64 into `messages.parts`, so image-heavy sessions stored multi-megabyte blobs inline in SQLite on every add/update/rewrite. That made saves, reloads, and resume/replay paths much slower and more memory-hungry. Persisting only lightweight metadata when a durable file path already exists keeps the database small without breaking multimodal replay after resume.

### DIFF
--- a/internal/llm/anthropic.go
+++ b/internal/llm/anthropic.go
@@ -656,13 +656,14 @@ func buildAnthropicBlocks(parts []Part, allowToolUse bool) []anthropic.ContentBl
 				blocks = append(blocks, anthropic.NewTextBlock(part.Text))
 			}
 		case PartImage:
-			if part.ImageData != nil {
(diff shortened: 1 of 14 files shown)


SamSaffron:

PR #680, opened 2026-05-18
fix: Uploaded images are fully decoded, saved, optionally resized, re-encoded, and kept inline before inference

sam-saffron-jarvis:
## What

- stream base64 uploads directly to the uploads directory instead of decoding the whole image into an intermediate raw buffer first
- only decode/resize large inline images after the original upload has been saved, using the saved file as the resize source
- factor upload temp-file creation so both raw-byte and streamed-base64 saves share the same atomic write path
- add a regression test to ensure invalid inline image uploads do not leave temp files behind

## Why

The previous image-upload hot path eagerly decoded every inline image, kept a full raw copy in memory, wrote that copy to disk, and then sometimes decoded/resized/re-encoded another copy before inference even started. For large or multi-image requests this caused avoidable CPU work, extra memory pressure, and slower upload/start latency.

This change preserves existing behavior — original uploads are still saved to disk and oversized inline payloads are still compressed for providers — while eliminating the full decoded in-memory copy for normal uploads and avoiding an extra raw-image buffer on the save path for large ones.

### DIFF
--- a/cmd/image_resize.go
+++ b/cmd/image_resize.go
@@ -8,6 +8,7 @@ import (
 	_ "image/png"
 	"log"
 	"math"
+	"os"
 
 	"golang.org/x/image/draw"
 	_ "golang.org/x/image/webp"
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #681, opened 2026-05-18
perf: reuse OpenAI SSE event buffers

sam-saffron-jarvis:
## What

Reuse an owned SSE event data buffer in the OpenAI-compatible streaming parser instead of allocating a fresh `[]byte` for every `data:` event.

The streaming hot path now constructs one `sseEventReader` per HTTP response and reuses its data buffer for each event. The parser still copies `bufio.Reader` slices before continuing to read the event separator, so the returned data is safe until the next event read.

## Why

A broad benchmark/profile pass showed SSE parsing as a small but real allocation hotspot in token streaming. OpenAI-compatible streams commonly deliver one JSON payload per SSE `data:` event, so per-event allocations add avoidable GC pressure and latency during long responses.

## Evidence

New benchmark over a 128-event chat-completions SSE payload:

Before:

```text
BenchmarkReadSSEEventBytesChatCompletions-32  156698-178912  7086-7227 ns/op  12328 B/op  131 allocs/op
```

After:

```text
BenchmarkReadSSEEventBytesChatCompletions-32  242428-250070  4775-4989 ns/op   4288 B/op    4 allocs/op
```

Representative improvement: about 31% lower parser time, 65% fewer allocated bytes, and 127 fewer allocations per 128-event payload.

## Verification

```text
go test ./internal/llm -run 'TestReadSSEEvent|TestSSEEventReader|TestOpenAICompatStream_AllowsLargeSSEDataLines'
go test ./internal/llm -run 'TestReadSSEEvent|TestOpenAICompatStream_AllowsLargeSSEDataLines' -bench 'BenchmarkReadSSEEvent(Bytes)?ChatCompletions' -benchmem -count=5
go build ./...
go test ./...
```

### DIFF
--- a/internal/llm/openai_compat.go
+++ b/internal/llm/openai_compat.go
@@ -437,14 +437,40 @@ func readSSEEvent(reader *bufio.Reader) (eventType, data string, eof bool, err e
 	}
 }
 
+type sseEventReader struct {
+	reader *bufio.Reader
+	data   []byte
+}
+
+func newSSEEventReader(reader *bufio.Reader) *sseEventReader {
+	return &sseEventReader{reader: reader}
+}
+
 func readSSEEventBytes(reader *bufio.Reader) (eventType string, data []byte, eof bool, err error) {
+	return newSSEEventReader(reader).readEventBytes()
+}
+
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #682, opened 2026-05-18
perf: defer skill metadata catalog loading

sam-saffron-jarvis:
## What

Defer the expensive skills catalog scan during `skills.NewSetup`.

`NewSetup` now uses `Registry.HasAnySkill()` as a cheap availability probe and leaves the full `Registry.List()` catalog load/parsing until `EnsurePromptMetadata()` is actually called. This keeps the activation/search tools available without eagerly parsing every `SKILL.md` on startup paths where metadata injection is skipped (for example when AGENTS.md already provides skills markup).

## Why

The startup skills setup path was still doing a full metadata catalog load just to decide whether skills existed. With large skill directories this repeats disk reads/YAML parsing before the prompt metadata is known to be needed.

## Evidence

Added `BenchmarkNewSetupWithoutPromptMetadata` with 200 project skills.

Before:

```text
BenchmarkNewSetupWithoutPromptMetadata-32  262-289 iters  4.12-4.30 ms/op  ~5.82 MB/op  31,743-31,744 allocs/op
```

After:

```text
BenchmarkNewSetupWithoutPromptMetadata-32  22,485-24,496 iters  49.7-53.5 us/op  ~41.6 KB/op  508 allocs/op
```

Also verified:

```text
go build ./...
go test ./...
go test ./internal/skills -run '^$' -bench 'BenchmarkNewSetupWithoutPromptMetadata|BenchmarkRegistryListRepeated|BenchmarkRegistryListCold' -benchmem -count=5
```

### DIFF
--- a/internal/skills/setup.go
+++ b/internal/skills/setup.go
@@ -21,7 +21,6 @@ type Setup struct {
 	alwaysEnabled        []string
 	metadataBudgetTokens int
 	maxVisibleSkills     int
-	preloadedSkills      []*Skill // Primed startup catalog reused for first prompt metadata build
 
 	metadataOnce sync.Once
 	metadataErr  error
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #689, opened 2026-05-19
fix: enforce output tool finalization

janzenisaac:
## Summary\n- make configured output tools contractual: natural completion now triggers one constrained finalization turn with only the required output tool available\n- return a typed nonzero error if the output tool still is not called instead of falling back to assistant prose\n- treat empty-string output-tool values as real calls and run on_complete with them\n\n## Tests\n- `go test ./...`


janzenisaac:
Accidental

PR #691, opened 2026-05-20
fix: Memory mining paginates long sessions with OFFSET, making reads degrade toward O(n²)

sam-saffron-jarvis:
## What

Change memory mining to load session messages from the last mined sequence with `GetMessagesFrom` instead of repeatedly calling `GetMessages(..., OFFSET ...)` as the offset grows.

Also add a regression test that verifies mining uses sequence-based reads rather than offset pagination.

## Why

The previous implementation re-ran `ORDER BY sequence ASC LIMIT ? OFFSET ?` for every batch and incremented the offset one message at a time. On long sessions that makes SQLite skip more and more rows on each batch, so catch-up mining work gets progressively slower and wastes database effort.

Using a sequence seek lets SQLite use the `(session_id, sequence)` index directly and keeps reads linear as sessions get larger.

### DIFF
--- a/cmd/memory_mine.go
+++ b/cmd/memory_mine.go
@@ -623,21 +623,24 @@ func loadMessagesForMining(ctx context.Context, store session.Store, candidate m
 	all := make([]session.Message, 0, memoryMineBatchSize)
 	result := memoryMineLoadResult{}
 
-	for {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #692, opened 2026-05-20
fix: Vector memory search scores every stored embedding blob on every query

sam-saffron-jarvis:
## What

- bound `VectorSearch` to a generous recent embedding window instead of rescoring every embedding blob for the provider/model on every query
- order that window by `embedded_at DESC` and add a matching composite index so SQLite can satisfy the prefilter without a temp sort
- keep exact cosine scoring and top-N heap selection inside that bounded candidate set, with the existing fragment fetch-after-score behavior
- add a query-plan test that locks in use of the recent-embedding index

## Why

Vector memory search is used interactively and currently rescored every stored embedding for the selected model as the memory DB grows. On long-lived installs that makes retrieval steadily slower and more expensive even though callers usually only need the top ~24 candidates.

This change caps the amount of blob decoding and cosine work per query while preserving exact scoring within a large recent candidate window, which keeps memory retrieval responsive as the embedding corpus grows.

### DIFF
--- a/internal/memory/store.go
+++ b/internal/memory/store.go
@@ -132,6 +132,7 @@ CREATE TABLE IF NOT EXISTS memory_embeddings (
 );
 
 CREATE INDEX IF NOT EXISTS idx_memory_embeddings_provider_model_dimensions ON memory_embeddings(provider, model, dimensions);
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #693, opened 2026-05-20
fix: Vector memory search linearly scans and scores every embedding blob on every query

sam-saffron-jarvis:
## What

- add lightweight vector-search metadata columns for embeddings (prefix dimensions plus tail/full norms) and populate them on upsert
- lazily backfill that metadata for older/directly-inserted rows before search
- change `VectorSearch` to rank candidates by a SQL-computed upper bound, then fetch and score exact embedding blobs in batches until the exact top-k is proven
- add a covering index for the new SQL candidate scan and update the query-plan test accordingly
- make in-memory SQLite stores use a shared-cache DSN so multi-query vector-search tests continue to work

## Why

`VectorSearch` previously loaded and scored every matching embedding blob in Go on every query, so memory retrieval degraded toward a full-table vector scan as the embedding corpus grew.

This keeps the search exact, but moves candidate pruning into SQL using small scalar metadata instead of full blobs. In practice that sharply reduces blob reads, decoding work, and per-query CPU for larger long-lived memory databases while preserving current behavior and compatibility with legacy rows.

### DIFF
--- a/internal/memory/store.go
+++ b/internal/memory/store.go
@@ -122,16 +122,42 @@ CREATE VIRTUAL TABLE IF NOT EXISTS memory_fts USING fts5(
 );
 
 CREATE TABLE IF NOT EXISTS memory_embeddings (
-    fragment_id TEXT NOT NULL REFERENCES memory_fragments(id) ON DELETE CASCADE,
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #694, opened 2026-05-20
fix: Memory mining paginates session messages with OFFSET, causing reads to degrade toward O(n²)

sam-saffron-jarvis:
## What

- switch memory mining from OFFSET-based pagination to sequence-based reads via `GetMessagesFrom(..., fromSeq, limit)`
- teach the SQLite session store to apply an optional `LIMIT` when seeking from a sequence cursor
- update `memory freshen` to reuse the same sequence-based read path
- add tests covering limited `GetMessagesFrom` reads and verifying memory mining uses sequence pagination

## Why

`memory mine` walked long sessions by repeatedly calling `GetMessages(... LIMIT ? OFFSET ?)` with larger and larger offsets. On SQLite that makes each successive batch more expensive because the database has to skip more rows before returning the next page, so mining old or high-message-count sessions slows down the deeper it gets into the transcript.

Using the session message sequence as a cursor lets SQLite seek directly into the `(session_id, sequence)` index and continue forward, avoiding the repeated OFFSET scans while preserving the existing mined-offset/state behavior.

### DIFF
--- a/cmd/memory_freshen.go
+++ b/cmd/memory_freshen.go
@@ -125,7 +125,7 @@ func runMemoryUpdateRecent(cmd *cobra.Command, args []string) error {
 			return err
 		}
 
-		messages, err := sessStore.GetMessages(ctx, sess.ID, 0, startOffset)
+		messages, err := sessStore.GetMessagesFrom(ctx, sess.ID, startOffset, 0)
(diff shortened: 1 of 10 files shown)


SamSaffron:

PR #695, opened 2026-05-20
fix: materialize local image paths

janzenisaac:
## Summary
- detect local image paths in text content and materialize readable files into the uploads directory
- attach materialized images as normal image parts with inline LLM payloads and durable ImagePath copies
- return a clear error for short-lived macOS TemporaryItems screenshot paths that are already unavailable or unreadable

## Tests
- go test ./...

### DIFF
--- a/cmd/serve_protocol.go
+++ b/cmd/serve_protocol.go
@@ -17,6 +17,7 @@ import (
 	"strconv"
 	"strings"
 	"time"
+	"unicode"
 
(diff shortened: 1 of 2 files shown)


janzenisaac:
Closing after local testing showed the failure was Docker/macOS privacy access to /private/var/folders/TemporaryItems, not a term-llm server-side materialization issue.

PR #697, opened 2026-05-20
fix: unblock provider stream close during SSE reads

sam-saffron-jarvis:
## What

- Add a cancellable stream close hook to `newEventStream` so providers can unblock resources that were opened before the stream goroutine starts.
- Use that hook to close HTTP response bodies for OpenAI-compatible, Responses API HTTP, Venice, xAI, and Ollama streaming adapters when `Stream.Close()` is called.
- Suppress expected close-induced read errors after the stream context has already been canceled.
- Add a deterministic OpenAI-compatible regression test using `io.Pipe` to verify `stream.Close()` returns while the provider goroutine is blocked waiting for more SSE bytes.

## Why

Some HTTP/SSE adapters perform the initial request synchronously, then hand the already-open `resp.Body` to `newEventStream`. `Stream.Close()` previously only canceled the child stream context; it did not cancel the original HTTP request context or close the response body. If the stream goroutine was blocked in a body read waiting for more SSE bytes, `Close()` could wait indefinitely, adding avoidable latency/hangs during cancellation or early stream shutdown.

## Evidence

- Before the fix, the new regression scenario timed out after 500ms: `stream.Close() blocked while the provider goroutine was waiting for more SSE bytes`.
- After the fix: `go test ./internal/llm -run '^TestOpenAICompatStream_CloseUnblocksPendingRead$' -count=5` passes in 0.012s.
- Full verification: `go build ./...` and `go test ./...` pass.

### DIFF
--- a/internal/llm/ollama.go
+++ b/internal/llm/ollama.go
@@ -340,7 +340,7 @@ func (p *OllamaProvider) Stream(ctx context.Context, req Request) (Stream, error
 		return nil, fmt.Errorf("Ollama API error (status %d): %s", resp.StatusCode, string(raw))
 	}
 
-	return newEventStream(ctx, func(ctx context.Context, send eventSender) error {
+	return newEventStreamWithCancelHook(ctx, func() { _ = resp.Body.Close() }, func(ctx context.Context, send eventSender) error {
 		defer resp.Body.Close()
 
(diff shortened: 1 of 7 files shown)


SamSaffron:

PR #700, opened 2026-05-21
Add NEAR AI Cloud provider

PierreLeGuen:
## Summary

Adds NEAR AI Cloud as a built-in OpenAI-compatible provider.

- Adds the `nearai` provider type using `NEARAI_API_KEY` and `https://cloud-api.near.ai/v1`.
- Wires the provider into config defaults, provider discovery, the setup wizard, model listing, and documentation.
- Adds NEAR AI Cloud model catalog support via `/model/list`, including chat-model filtering, context windows, and token pricing metadata.

## Implementation Notes

This follows the repo's existing built-in OpenAI-compatible provider pattern. Chat completions and tool calls use the shared OpenAI-compatible transport, while model discovery uses NEAR AI Cloud's public catalog endpoint.

The model-listing path filters non-chat catalog entries such as image, embedding, audio, reranker, and privacy-filter models so the LLM provider surface only shows chat-capable models.

## Testing

- `gofmt -l .`
- `go build ./...`
- `go test ./...`
- `term-llm models --provider nearai --json`

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ If you already have a provider key:
 
 ```bash
 export ANTHROPIC_API_KEY=your-key
-# or OPENAI_API_KEY / GEMINI_API_KEY / OPENROUTER_API_KEY / XAI_API_KEY / SAMBANOVA_API_KEY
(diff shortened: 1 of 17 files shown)


SamSaffron:
Thanks Pierre, looks tidy

PR #701, opened 2026-05-21
fix: Stream consumption is intentionally stalled to one event per smooth-tick, causing upstream backpressure and delayed tool/done events

sam-saffron-jarvis:
## What

Stop deferring the next chat stream read behind a pending smooth tick. The chat TUI now keeps listening for the next provider event immediately after handling any non-terminal stream event, while still using the smooth buffer/tick path to coalesce visible text rendering. Updated the regression test to assert that streamed text keeps both the smooth tick and the follow-up stream read scheduled.

## Why

Deferring reads to one per smooth tick capped stream consumption at roughly one event every 16ms whenever providers emitted lots of tiny text deltas. That let backlog accumulate in the buffered adapter channel, eventually pushed back on the stream goroutine itself, and delayed tool start/end visibility plus final done transitions. Continuing to dequeue immediately removes that backpressure without giving up frame-paced rendering.

### DIFF
--- a/internal/tui/chat/chat.go
+++ b/internal/tui/chat/chat.go
@@ -1930,15 +1930,11 @@ func (m *Model) Update(msg tea.Msg) (tea.Model, tea.Cmd) {
 		}
 
 		// Continue listening for more events unless we're done or got an error.
-		// When a smooth tick is already pending for streamed text, defer the next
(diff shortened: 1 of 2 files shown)


SamSaffron:
Superseded; integrating cleaned-up changes directly.

PR #702, opened 2026-05-21
fix: Streaming partial markdown reparses and re-renders the entire in-progress segment on every frame

sam-saffron-jarvis:
## What

- stop full-document markdown re-renders on every flowing partial update once a block preview is established
- cache the stable rendered prefix before the active partial block and reuse it for later token updates in the same block
- add a regression test that verifies flowing partial updates stop calling the renderer with committed markdown on every frame

## Why

`TextSegmentRenderer` uses flowing partial rendering on the Bubble Tea path, and every token update was rebuilding a synthetic markdown document from all committed content plus the current safe partial block. That made streaming work scale with already-rendered history and caused visible jank as responses grew.

This keeps the first flowing partial render exact, then reuses its stable prefix so subsequent updates only re-render the active block instead of reparsing committed markdown on every frame.

### DIFF
--- a/internal/ui/streaming/partial.go
+++ b/internal/ui/streaming/partial.go
@@ -7,10 +7,21 @@ import (
 
 // partialState tracks the state of partial block rendering for re-rendering.
 type partialState struct {
-	safeMarkdown string // Markdown content up to incomplete syntax
-	safeRendered string // Rendered output currently displayed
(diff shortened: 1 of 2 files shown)


SamSaffron:
Superseded; integrating cleaned-up changes directly.

PR #703, opened 2026-05-21
fix: Streaming partial markdown reparses and re-renders the entire in-progress segment on every smooth tick

sam-saffron-jarvis:
## What

- avoid re-rendering the full committed markdown snapshot on every flowing partial-preview tick
- reuse the already-rendered committed prefix after the first partial preview for a block, and only re-render the current partial block on subsequent updates
- add regression coverage for both the reduced render behavior and parity with full rendering across partial updates

## Why

Streaming assistant responses were reparsing and re-rendering all committed markdown plus the active partial block on every smooth tick. As segments grew, each token frame became progressively more expensive and streaming could visibly stutter. Reusing the stable rendered prefix keeps the common path bounded to the active partial block while preserving output correctness.


SamSaffron:

PR #704, opened 2026-05-21
fix: The UI only dequeues one stream event per frame, creating avoidable backlog and upstream backpressure

sam-saffron-jarvis:
## What

- keep scheduling the next chat stream read immediately after each non-terminal stream event, even while smooth text rendering is already frame-paced
- remove the now-unused deferred stream-read path from the chat TUI
- update the chat perf test to assert we keep both the smooth tick and the immediate follow-up read scheduled

## Why

The previous logic only dequeued one stream event per smooth-render frame when text deltas were arriving quickly. That let the bounded stream adapter channel back up and apply avoidable upstream backpressure, which added latency to token display and delayed follow-up tool/done events. Draining the stream continuously preserves the existing coalesced rendering behavior without making providers wait on the UI.

### DIFF
--- a/internal/tui/chat/chat.go
+++ b/internal/tui/chat/chat.go
@@ -88,7 +88,6 @@ type Model struct {
 	// Smooth text buffer for 60fps rendering
 	smoothBuffer            *ui.SmoothBuffer
 	smoothTickPending       bool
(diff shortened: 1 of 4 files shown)


SamSaffron:

PR #705, opened 2026-05-21
optimize session snapshot replacement

sam-saffron-jarvis:
## What

Optimize `SQLiteStore.ReplaceMessages` so full-session snapshot persistence keeps an unchanged message prefix and only deletes/reinserts the changed suffix.

The fallback path still performs a full rewrite if legacy/corrupt rows have duplicate or negative sequence numbers, preserving the complete-history replacement contract.

## Why

Serve/web persistence commonly rebuilds a fresh full snapshot for a session where most prior messages are unchanged. The old implementation deleted every message and reinserted the entire transcript, churning FTS triggers, `message_count` triggers, row IDs, and SQLite writes even when only the tail changed.

This keeps stable history rows in place and avoids reindexing them.

## Evidence

Focused benchmark on an unchanged 120-message snapshot:

- Before: `BenchmarkSQLiteStoreReplaceMessagesLargeSnapshot` ~3.55-3.68 ms/op, ~132 KB/op
- After: `BenchmarkSQLiteStoreReplaceMessagesLargeSnapshot` ~245-249 µs/op, ~131 KB/op

Verification:

- `go test ./internal/session`
- `go test ./internal/session -run '^$' -bench 'BenchmarkSQLiteStoreReplaceMessagesLargeSnapshot' -benchmem -count=3`
- `go test ./...`
- `go build ./...`

### DIFF
--- a/internal/session/sqlite.go
+++ b/internal/session/sqlite.go
@@ -1648,10 +1648,17 @@ func (s *SQLiteStore) ClearCompactionBoundary(ctx context.Context, id string) er
 	})
 }
 
-// ReplaceMessages deletes all existing messages for the session and inserts
-// the new set in a single transaction. Because the replacement snapshot becomes
-// the complete persisted history, any previous compaction boundary is cleared.
+// ReplaceMessages reconciles the complete persisted history for a session.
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #706, opened 2026-05-21
perf: cache tool schema JSON for adapters

sam-saffron-jarvis:
## What

Cache JSON-encoded immutable `ToolSpec.Schema` maps by schema identity and reuse the cached `json.RawMessage` when building tool definitions for OpenAI-compatible chat-completions adapters and Ollama.

This avoids re-marshalling every tool schema on every streaming turn while preserving the existing per-tool error context if encoding fails.

## Why

Provider requests with many registered tools rebuild the same function/tool parameter JSON every agentic turn. `ToolSpec.Schema` is documented as immutable after registration, and `ToolRegistry` intentionally shares schema maps across repeated `AllSpecs()` calls, so this serialization work is redundant in normal usage.

## Evidence

Added `BenchmarkBuildCompatTools` to model repeated turns with 128 registered tools.

Before:

```text
BenchmarkBuildCompatTools-32    306996-313175 ns/op    233096-233104 B/op    6017 allocs/op
```

After:

```text
BenchmarkBuildCompatTools-32      7450-7864 ns/op       18688 B/op            257 allocs/op
```

Also added a unit test proving the same immutable schema is marshaled once and reused across both the OpenAI-compatible and Ollama tool builders.

## Verification

```text
go test ./internal/llm -run='TestToolSchemaJSONCacheReusesImmutableSchema'
go test ./internal/llm -run=^$ -bench=BenchmarkBuildCompatTools -benchmem -count=5
go build ./...
go test ./...
```

### DIFF
--- a/internal/llm/ollama.go
+++ b/internal/llm/ollama.go
@@ -229,7 +229,7 @@ func buildOllamaTools(specs []ToolSpec) ([]ollamaTool, error) {
 	}
 	tools := make([]ollamaTool, 0, len(specs))
 	for _, spec := range specs {
-		schema, err := json.Marshal(spec.Schema)
+		schema, err := cachedToolSchemaJSON(spec.Schema)
 		if err != nil {
 			return nil, fmt.Errorf("marshal tool schema %s: %w", spec.Name, err)
 		}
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #708, opened 2026-05-23
fix: TUI chat streams ignore the model’s root cancellation context, so quit/signal can leave provider and tool work running detached

sam-saffron-jarvis:
## What changed

- Changed TUI chat streaming to derive its cancellable run context from `m.rootContext()` instead of `context.Background()`.
- Applied the same root-context derivation to `/compact` and LLM-backed `/handover` helper operations.
- Added targeted tests covering stream, compaction, and handover cancellation propagation from the model root context.

## Why this is high-value

The chat model already receives a signal-aware parent context from `cmd/chat.go`, but these long-running operations were ignoring it and starting from `context.Background()`. That meant quitting the TUI or receiving SIGINT/SIGTERM could tear down the UI while the provider request, MCP activity, or tool/subprocess work kept running detached until its own timeout or completion.

This fix makes outer shutdown reliably cancel in-flight provider and tool work, which is especially important for providers and tools that spawn subprocesses or hold external resources.

## Validation

- Verified the issue in the current code: `startStream`, `/compact`, and LLM-backed `/handover` were all using `context.WithCancel(context.Background())` despite the existing root context plumbing.
- Ran `gofmt -w internal/tui/chat/streaming.go internal/tui/chat/commands.go internal/tui/chat/root_context_test.go`
- Ran `go build ./...`
- Ran `go test ./...`
- Ran `git diff --stat` and `git diff --check`

### DIFF
--- a/internal/tui/chat/commands.go
+++ b/internal/tui/chat/commands.go
@@ -1755,7 +1755,7 @@ func (m *Model) cmdCompress() (tea.Model, tea.Cmd) {
 
 	return m, tea.Batch(
 		func() tea.Msg {
-			ctx, cancel := context.WithCancel(context.Background())
+			ctx, cancel := context.WithCancel(m.rootContext())
 			m.streamCancelFunc = cancel
 			result, err := llm.Compact(ctx, provider, model, systemPrompt, llmMessages, compactConfig)
 			return compactDoneMsg{result: result, err: err}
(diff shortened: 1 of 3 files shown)


SamSaffron:
Closing in favor of implementing the corrected version directly on main/working tree, including the remaining cancellation path.

PR #709, opened 2026-05-23
fix: `recent.md` maintenance does unlocked read-modify-write overwrites, so overlapping runs can silently discard memory updates

sam-saffron-jarvis:
## What changed

- Added a shared `recent.md` update helper that acquires a per-file advisory lock before reloading `recent.md`, holds it across the LLM-based read/modify/write flow, and releases it after the write and related metadata updates complete.
- Switched both `memory update-recent` and `memory promote` to use that helper so overlapping maintenance runs no longer compute from the same stale `recent.md` snapshot and then blindly overwrite each other.
- Replaced direct `os.WriteFile` calls for `recent.md` with temp-file + rename writes so readers see either the old file or the fully written new file, never a partially written one.
- Added tests covering lock serialization and atomic `recent.md` replacement behavior.

## Why this is high-value

`recent.md` is Jarvis's current working memory. Before this change, `update-recent` and `promote` both did unlocked read-LLM-write cycles against the same file, so two overlapping runs could each read the same old content and the later writer would silently discard the earlier run's newly computed memory updates. Because these commands are natural scheduled-job/manual-maintenance paths, overlap is realistic and the failure mode is persistent memory loss.

This fix prevents that exact failure by serializing writers per `recent.md` path and making the final file replacement atomic.

## Validation

- `gofmt -w cmd/memory_freshen.go cmd/memory_promote.go cmd/memory_recent_file.go cmd/memory_recent_lock_unix.go cmd/memory_recent_lock_windows.go cmd/memory_recent_file_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/cmd/memory_freshen.go
+++ b/cmd/memory_freshen.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"io"
 	"os"
-	"path/filepath"
 	"sort"
 	"strconv"
 	"strings"
@@ -164,10 +163,6 @@ func runMemoryUpdateRecent(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return err
 	}
-	existingRecent, err := loadRecentContent(recentPath)
-	if err != nil {
-		return err
-	}
 
 	fragmentsText, err := loadRecentFragmentsText(ctx, store, agentName, memoryUpdateRecentFragmentChars)
 	if err != nil {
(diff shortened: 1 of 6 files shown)


SamSaffron:
Closing in favor of implementing the corrected version directly on main/working tree with full transaction locking and regression coverage.

PR #710, opened 2026-05-23
fix: Durable response ID lookups load and deserialize entire session histories on hot response paths

sam-saffron-jarvis:
## What changed

- Added `GetLatestVisibleMessageID` to the session store interface.
- Implemented it in `internal/session/sqlite.go` as a targeted single-row query over the `messages` table:
  - filters to visible continuation roles (`user`, `assistant`)
  - orders by newest sequence
  - returns only the latest message ID
- Replaced the full-history `GetMessages(..., 0, 0)` calls in `cmd/serve_responses_durable.go` with the new targeted lookup for:
  - durable `previous_response_id` validation
  - latest durable response ID reporting for a session
- Added a SQLite test covering the behavior that non-visible tail rows (for example `tool` or `event`) are skipped.
- Updated the no-op, logging, and test store implementations to satisfy the new store interface.

## Why this is high-value

These durable response ID helpers sit on hot Responses/web paths, including completion finalization, continuation validation, and ask-user state reporting. Before this change they loaded the entire persisted transcript for the session and deserialized every message `parts` JSON blob just to answer a tail question.

This fix reduces that work from:

- O(n) row reads over the whole session history
- O(n) JSON deserialization/allocation for message parts

…to a single small SQL lookup that returns one ID.

That directly cuts latency and allocation cost for long-lived persisted sessions, which are exactly the sessions most likely to hit this path repeatedly.

## Validation

- Verified the issue in current code by inspecting `cmd/serve_responses_durable.go` and `internal/session/sqlite.go`.
- `gofmt -w` on changed files
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/cmd/serve_responses_durable.go
+++ b/cmd/serve_responses_durable.go
@@ -83,15 +83,14 @@ func (s *serveServer) resolveDurablePreviousResponseID(ctx context.Context, prev
 	if headerSessionID != "" && headerSessionID != msg.SessionID {
 		return durablePreviousResponseResolution{}, http.StatusConflict, fmt.Sprintf("session_id %q conflicts with previous_response_id session %q", headerSessionID, msg.SessionID)
 	}
-	msgs, err := s.store.GetMessages(ctx, msg.SessionID, 0, 0)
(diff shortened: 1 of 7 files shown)


SamSaffron:

PR #711, opened 2026-05-23
fix: Durable response-id lookups load and JSON-decode the entire session transcript on hot response paths

sam-saffron-jarvis:
## What changed

- Added a targeted SQLite lookup for the latest visible persisted message in a session: `GetLatestVisibleMessageID`, which reads a single `user`/`assistant` row ordered by `sequence DESC LIMIT 1`.
- Updated durable response-id resolution to use that lookup instead of loading and JSON-decoding the entire session transcript just to validate whether a `previous_response_id` is the current tail.
- Updated session message pagination responses to compute `lastResponseId` from the full session tail, not from the current page contents.
- Covered the change with tests for both the SQLite lookup semantics and paginated session message responses.

## Why this is high-value

- This removes unnecessary work from a hot persisted-session path: long sessions no longer require `GetMessages(..., 0, 0)` plus per-row JSON decode just to find the last visible response id.
- It fixes a user-visible correctness issue in paginated web session reloads: the first page of `/v1/sessions/:id/messages` could previously return a stale `lastResponseId` from the truncated page, causing continuation conflicts even though the session itself was valid.
- The fix is narrowly scoped and preserves existing behavior everywhere else.

## Validation

- `gofmt -w cmd/serve_responses_durable.go cmd/serve_handlers.go internal/session/sqlite.go internal/session/logger.go internal/session/sqlite_test.go cmd/serve_test.go`
- `go build ./...`
- `go test ./...`
- Added/updated tests:
  - `TestSQLiteStoreGetLatestVisibleMessageIDSkipsInvisibleTail`
  - `TestHandleSessionMessages_DefaultPagination` now asserts `lastResponseId` stays pinned to the latest visible message across pages

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -1297,7 +1297,7 @@ func (s *serveServer) handleSessionByID(w http.ResponseWriter, r *http.Request) 
 		result = append(result, entry)
 	}
 
-	resp := messagesResponse{LastResponseID: latestDurableResponseID(msgs), Messages: result, HasMore: hasMore}
+	resp := messagesResponse{LastResponseID: s.latestDurableResponseIDForSession(r.Context(), sessionID), Messages: result, HasMore: hasMore}
 	if hasMore {
 		resp.NextOffset = nextOffset
 	}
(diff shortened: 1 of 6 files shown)


SamSaffron:

PR #712, opened 2026-05-23
fix: MCP server startup can hang indefinitely and leak stuck subprocesses because `Enable` depends on caller context

sam-saffron-jarvis:
## What changed

- Added a manager-owned startup timeout for MCP servers so `Enable` no longer depends on the caller passing a cancelable context.
- Stored per-server startup cancel functions in `Manager` and invoked them from `Disable`/`StopAll` so an in-flight startup can be interrupted.
- Ignored stale startup completions after a server has been disabled or replaced, which prevents canceled starts from flipping the server back to `failed`/`ready`.
- Added regression tests covering:
  - background-context startup timing out instead of hanging forever
  - `Disable` canceling an in-flight startup and leaving the server in `stopped`

## Why this is high-value

Today several TUI call sites use `context.Background()` when enabling/restarting MCP servers. A misbehaving stdio server can therefore hang forever inside `Client.Start`, keep the server stuck in `starting`, and block reliable cleanup because `Disable` had no manager-level way to cancel that startup.

This fix makes MCP startup self-bounded and recoverable from the manager layer, which is important for long-running interactive sessions where a broken MCP server should not require a full app restart.

## Validation

- `gofmt -w internal/mcp/manager.go internal/mcp/manager_test.go`
- `go build ./...`
- `go test ./...`
- Added `internal/mcp/manager_test.go` regression coverage for both timeout and disable-cancel behavior.

### DIFF
--- a/internal/mcp/manager.go
+++ b/internal/mcp/manager.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"sort"
 	"sync"
+	"time"
 
 	"github.com/samsaffron/term-llm/internal/llm"
 )
@@ -19,6 +20,8 @@ const (
 	StatusReady    ServerStatus = "ready"
 	StatusFailed   ServerStatus = "failed"
 )
+
+var mcpStartupTimeout = 30 * time.Second
 
 // ServerState holds the state of a managed MCP server.
 type ServerState struct {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #713, opened 2026-05-24
fix: Interrupted Telegram replies are only kept in RAM and are dropped from persisted session history

sam-saffron-jarvis:
## What changed

- Persist the partial assistant reply in the Telegram interrupt path when streaming was canceled before any assistant message had been durably written.
- Use a short timeout with `context.WithoutCancel(...)` for that best-effort write so the interrupt itself does not cancel the save.
- Keep in-memory history aligned with the persisted fallback logic by treating "assistant already captured" separately from "any produced messages exist".
- Add a regression test covering an interrupted plain-text Telegram reply and asserting the partial assistant text is stored with interrupted session status.

## Why this is high-value

Interrupting an in-flight Telegram reply is a normal user action. Before this change, the user could see a partial answer plus an interrupted marker in Telegram, but that assistant text was never saved to the session store if streaming stopped before completion callbacks fired. After a restart or when restoring history from disk, the visible assistant output was lost. This change closes that real data-loss path with a minimal, targeted fallback.

## Validation

- `go build ./...`
- `go test ./...`
- Added `TestStreamReply_PersistsInterruptedPartialAssistantReply`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -1521,6 +1521,13 @@ loop:
 		producedMu.Lock()
 		producedSnapshot := append([]llm.Message(nil), produced...)
 		producedMu.Unlock()
+		assistantPersisted := false
+		for _, msg := range producedSnapshot {
+			if msg.Role == llm.RoleAssistant {
+				assistantPersisted = true
+				break
+			}
+		}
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #714, opened 2026-05-24
fix: Response-run replay disables compaction forever after approvals/interjections and can retain every streamed token event

sam-saffron-jarvis:
## What changed

- Kept response-run event compaction enabled even after `response.ask_user.prompt`, `response.approval.prompt`, and `response.interjection` so the raw replay log still respects `maxRetainedEvents`.
- Preserved interactive recovery state separately from the raw event buffer:
  - ask-user and approval prompt payloads are now retained in `recovery.events`
  - interjections are now captured in `recovery.messages` as a user message with `interruptState: "interject"`, which also correctly splits assistant output across the interruption
- Added focused tests covering:
  - compaction still happening after interactive prompts / `disableCompaction()` call sites
  - prompt-only recovery snapshots retaining interactive events
  - interjections surviving snapshot recovery and splitting assistant messages correctly

## Why this is high-value

Before this change, any approval, ask-user prompt, or interjection permanently disabled response-run compaction. After that point, long runs retained every `response.output_text.delta` event in memory until run cleanup, so per-run heap usage and replay work grew linearly with streamed output size.

This fix restores the bounded 2048-event replay window for interactive runs while still preserving the interactive state needed for snapshot-based recovery. That reduces retained allocations, GC pressure, and reconnect replay cost for the long-running serve process, especially on agentic runs that stream lots of tokens after an approval or interruption.

## Validation

- `go build ./...`
- `go test ./...`
- Added/ran targeted tests in `cmd/serve_response_runs_test.go` for the compaction and recovery behavior above.

### DIFF
--- a/cmd/serve_response_runs.go
+++ b/cmd/serve_response_runs.go
@@ -34,14 +34,20 @@ type responseRunRecoveryTool struct {
 }
 
 type responseRunRecoveryMessage struct {
-	ID       string
-	Role     string
-	Content  []byte
-	Created  int64
-	Tools    []responseRunRecoveryTool
-	Expanded bool
-	Status   string
-	Usage    map[string]any
+	ID             string
+	Role           string
+	Content        []byte
+	Created        int64
+	Tools          []responseRunRecoveryTool
+	Expanded       bool
+	Status         string
+	Usage          map[string]any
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #715, opened 2026-05-24
fix: Interrupted Telegram replies are shown to the user but not durably persisted

sam-saffron-jarvis:
## What changed

- Persist a best-effort assistant snapshot in the Telegram interrupt path when partial text was shown but no assistant message was captured by the normal callbacks yet.
- Use a short detached store context (`context.WithoutCancel` + timeout) for that final write so the user interrupt does not cancel the persistence attempt.
- Add a regression test covering an agentic Telegram stream that emits partial assistant text, gets interrupted, and must still restore the partial assistant reply from the DB.

## Why this is high-value

Interrupted Telegram replies are visible to the user immediately, so dropping that partial assistant turn from durable history creates a real state mismatch after restart or DB-backed restore. This is a common interactive flow: send a message, see part of the answer, then send a follow-up that interrupts it. Persisting the last visible assistant snapshot prevents silent conversation-history loss.

## Validation

- `go build ./...`
- `go test ./...`
- Added `TestStreamReply_InterruptPersistsPartialAssistantFallback`

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -801,6 +801,20 @@ func (m *telegramSessionMgr) runStoreOpWithTimeout(sessionID, op string, fn func
 	}
 }
 
+func (m *telegramSessionMgr) runStoreOpWithoutCancel(ctx context.Context, sessionID, op string, fn func(context.Context) error) {
+	if m.store == nil || fn == nil {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #721, opened 2026-05-26
fix: FTS session-browser refresh loads full transcripts for every hit

sam-saffron-jarvis:
## What changed

- extended `session.SearchResult` and `SQLiteStore.Search` to return the lightweight session metadata the browser already needs: mode, status, persisted `message_count`, session created time, and updated time
- updated the TUI sessions browser FTS refresh path to build rows directly from `Search(...)` metadata instead of calling `Get(...)` and `GetMessages(..., 0, 0)` for every deduplicated hit
- added:
  - a TUI unit test proving FTS refresh no longer calls `Get` or `GetMessages`
  - a SQLite search test covering the new metadata fields

## Why this is high-value

Before this change, one FTS refresh could do:
- 1 search query
- plus 1 `Get(...)` per unique session hit
- plus 1 `GetMessages(..., 0, 0)` per unique session hit

That meant loading full transcripts into memory just to compute `len(msgs)` for the browser list, even though `sessions.message_count` is already persisted specifically to avoid per-session counting scans.

This fix removes those per-hit full-history loads and replaces them with metadata already available from the search/listing layer. That reduces DB round-trips, transcript deserialization, allocations, and latency as session histories grow, which directly improves TUI search responsiveness.

## Validation

- `gofmt -w internal/session/types.go internal/session/sqlite.go internal/session/sqlite_test.go internal/tui/sessions/browse.go internal/tui/sessions/browse_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/session/sqlite.go
+++ b/internal/session/sqlite.go
@@ -1475,9 +1475,14 @@ func (s *SQLiteStore) Search(ctx context.Context, query string, limit int) ([]Se
 		return []SearchResult{}, nil
 	}
 
+	messageCountCol := "COALESCE(s.message_count, 0)"
+	if !s.hasMessageCount {
+		messageCountCol = "(SELECT COUNT(*) FROM messages WHERE session_id = s.id AND role IN ('user', 'assistant'))"
+	}
+
 	rows, err := s.db.QueryContext(ctx, `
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #722, opened 2026-05-26
fix: Telegram drops visible partial replies on stream timeout/error

sam-saffron-jarvis:
## What changed

- Added a small `salvagePartialHistory` helper in `internal/serve/telegram.go` that preserves visible partial assistant output plus callback-produced messages in `sess.history`.
- Reused the existing interrupt salvage behavior for the generic `streamErr != nil` path, including a best-effort fallback `AddMessage` persisted with `runStoreOpWithoutCancel` when partial assistant text was shown but not durably recorded yet.
- Added `TestStreamReply_PersistsErroredPartialAssistantReply` to cover the failure mode where Telegram already displayed partial text and the stream then errors.

## Why this is high-value

Telegram streaming can show partial assistant text to the user before a provider/network/watchdog failure ends the turn. Before this change, the interrupt path salvaged that visible text, but the generic error path only marked the session as `error` and returned. That meant the next turn could resume from history that did not match what the user already saw in Telegram.

This fix makes timeout/error handling preserve the same visible partial reply context that interrupts already preserve, which prevents real chat state loss after flaky streams.

## Validation

- Verified the issue in current code by reading the existing interrupt salvage path versus the generic `streamErr != nil` path in `internal/serve/telegram.go`.
- Added and ran targeted coverage:
  - `go test ./internal/serve -run 'TestStreamReply_(PersistsErroredPartialAssistantReply|PersistsInterruptedPartialAssistantReply)$'`
- Ran formatting and repo validation:
  - `go build ./...`
  - `go test ./...`
  - `git diff --check`

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -1467,6 +1467,38 @@ func (m *telegramSessionMgr) streamReply(ctx context.Context, bot botSender, ses
 		return true
 	}
 
+	salvagePartialHistory := func(persistCtx context.Context, fallbackOp string) {
+		textMu.Lock()
+		partial := textBuf.String()
+		textMu.Unlock()
+		producedMu.Lock()
+		producedSnapshot := append([]llm.Message(nil), produced...)
+		producedMu.Unlock()
+		if partial == "" && len(producedSnapshot) == 0 {
+			return
+		}
+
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #723, opened 2026-05-26
fix: Command-help overlay keeps the LLM stream running after the user quits

sam-saffron-jarvis:
## What changed

- switched `ShowCommandHelp` from `context.Background()` to `context.WithCancel(...)`
- extracted the background help-stream reader into `streamCommandHelp`
- cancel the help-stream context when the Bubble Tea program exits and wait for the stream goroutine to finish before returning
- suppress cancellation-shaped stream errors during shutdown so quitting the overlay does not surface a spurious error
- added a targeted test that proves the help stream stops promptly when its context is cancelled

## Why this is high-value

Dismissing the command-help overlay previously left the provider stream running in the background until the LLM request finished on its own. That leaked in-flight network/token work after the UI was gone, and repeated quick exits could accumulate hidden streams in the same CLI session.

This fix makes overlay exit deterministic: quitting now cancels the provider request and waits for cleanup before returning.

## Validation

- `gofmt -w internal/ui/postexec.go internal/ui/postexec_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/ui/postexec.go
+++ b/internal/ui/postexec.go
@@ -2,6 +2,7 @@ package ui
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"io"
 	"os"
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #729, opened 2026-05-27
feat(serveui): add cmd+k shortcut to open widgets modal

arpitjalan:
Open the widgets modal from anywhere in the web UI with a keyboard shortcut: cmd+k on macOS, ctrl+k on Linux/Windows. Pressing again closes the modal; esc already closes it via the existing handler.

Platform detection follows the standard $mod convention used by Notion, GitHub, and libraries like tinykeys: cmd on Mac, ctrl elsewhere - never both. Restricting to cmd-only on macOS preserves ctrl+k's emacs-style "kill to end of line" inside the chat textarea.

The binding is suppressed when the widgets sidebar button is hidden (no widgets loaded), and ignored if alt/shift modifiers are also held, so it can't fire accidentally.

### DIFF
--- a/internal/serveui/static/app-sidebar.js
+++ b/internal/serveui/static/app-sidebar.js
@@ -178,6 +178,21 @@ if (elements.widgetsModal) {
     }
   });
 }
+
+const isMac = /Mac|iPhone|iPad|iPod/.test(navigator.platform);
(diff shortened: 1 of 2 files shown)


SamSaffron:
thanks Arpit!

PR #730, opened 2026-05-28
feat(serve): honor TERM_LLM_SERVE_TOKEN env var

arpitjalan:
Let users pin a stable bearer token across `term-llm serve` runs by exporting TERM_LLM_SERVE_TOKEN. The web UI persists the token in localStorage, so a stable server-side value means users paste the token once per browser instead of after every restart. API clients and scripts with the token saved in config see the same benefit.

Precedence: --token > $TERM_LLM_SERVE_TOKEN > auto-generated. When neither flag nor env var is set, behavior is unchanged. Containerized `term-llm contain` workspaces are unaffected — their webui service already passes `--token` explicitly from each workspace's .env file, which keeps top precedence.

- Flag help advertises the env var.
- Startup banner annotates the token line with its source: "(from $TERM_LLM_SERVE_TOKEN)" when read from the env, or "(auto-generated; export TERM_LLM_SERVE_TOKEN to persist)" when freshly generated.
- New "Persist the token across restarts" subsection in the web UI guide with bash/zsh and fish examples.

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -9,6 +9,7 @@ import (
 	"io"
 	"log"
 	"net/http"
+	"os"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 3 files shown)


SamSaffron:
thanks Arpit!

PR #731, opened 2026-05-28
fix(tui): resolve symlinks on approved dirs in IsPathApproved

arpitjalan:
IsPathApproved canonicalised the file path with EvalSymlinks but only ran filepath.Abs on each approved directory. When the two sides differ through a symlink, the prefix check fails and the file is treated as unapproved.

In practice this breaks persistent approval on macOS: /tmp and /var/folders are symlinks to /private/tmp and /private/var, so every /file attachment from those paths re-prompts the user to approve a directory they already approved. The stored approval never sticks. Also affects symlinked project dirs on any OS (e.g. ~/code -> /Volumes/SSD/code).

TestCmdFileClearsComposerAfterAttach has been failing on macOS since it was added in efbd3f3 -- t.TempDir() returns a /var/folders path, which is exactly the case that trips this bug. CI on Linux didn't catch it.

Resolve symlinks on the approved dir too, falling back to the absolute path on error. Add TestApprovedDirsResolvesSymlinks with an explicit os.Symlink so the regression is caught on Linux as well.

### DIFF
--- a/internal/tui/chat/dirs.go
+++ b/internal/tui/chat/dirs.go
@@ -109,8 +109,17 @@ func (d *ApprovedDirs) IsPathApproved(path string) bool {
 			continue
 		}
 
+		// Resolve symlinks on the approved dir as well, so the comparison
+		// is apples-to-apples with realPath. Without this, an approved dir
+		// reached via a symlink (e.g. macOS /var -> /private/var) never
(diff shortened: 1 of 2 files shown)


SamSaffron:
I think this is a bit too late, let me try a fix that does this when we add the directories in , that way everything is canon everywhere.

PR #732, opened 2026-05-28
fix: Crash recovery requeues already-started jobs and can rerun side effects after restart

sam-saffron-jarvis:
## What changed

- Changed `jobsV2Manager.recoverRuns()` so startup recovery no longer rewrites `claimed` and `running` runs back to `queued`.
- Recovery now:
  - keeps `queued` runs untouched,
  - converts `cancel_requested` runs to `cancelled`, and
  - marks abandoned `claimed` / `running` runs as `failed` with `exit_reason=worker_lost` and error text `worker lost before run could finish`.
- Reused normal `finishRun()` handling during recovery so configured retry policy can schedule a fresh retry attempt instead of rerunning the interrupted attempt in place.
- Preserved any partial run output already stored on the abandoned row during recovery.
- Added tests covering:
  - `claimed` and `running` rows becoming terminal `failed` instead of requeued, and
  - retry scheduling for an abandoned run when retry policy allows another attempt.

## Why this is high-value

Before this change, a restart/crash would silently move persisted `claimed` and `running` rows back to `queued`, and workers would execute those same rows from scratch. Because jobs can run arbitrary commands, invoke tools, send notifications, or write files, that recovery path could duplicate real-world side effects.

This fix makes crash recovery safer: interrupted attempts are recorded as terminal worker-loss failures, and only an explicit retry policy can enqueue a new attempt.

## Validation

- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -56,13 +56,16 @@ const (
 )
 
 const (
-	exitReasonNatural   = "natural_completion" // agent finished normally
-	exitReasonMaxTurns  = "max_turns_exceeded" // hit the agentic loop turn limit
-	exitReasonTimeout   = "timeout"            // context deadline exceeded
-	exitReasonCancelled = "cancelled"          // context cancelled
-	exitReasonException = "exception"          // unhandled error
-	exitReasonEmpty     = "empty_response"     // succeeded but produced no output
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as requested; implementing the underlying recovery fix locally.

PR #733, opened 2026-05-28
fix: Crash recovery requeues already-started jobs, so restarts can duplicate side effects

sam-saffron-jarvis:
## What changed

- Changed `jobsV2Manager.recoverRuns()` so crash recovery no longer rewrites `claimed` and `running` runs back to `queued`.
- Recovery now:
  - leaves existing `queued` runs alone,
  - converts `cancel_requested` runs to `cancelled`, and
  - marks interrupted `claimed`/`running` runs as `failed` with a clear worker-lost error and `exit_reason=exception`.
- Extracted retry scheduling into a shared helper so recovered failed runs still get a fresh retry attempt when the job's retry policy allows it.
- Added a regression test covering interrupted `claimed` and `running` runs, verifying they are not silently requeued, preserve partial output, and create a new retry run instead.

## Why this is high-value

Before this change, a daemon restart or crash could cause an already-started job to run from scratch again, because recovery rewrote in-flight runs back to `queued`. That can duplicate side effects from program and LLM jobs: repeated notifications, repeated shell commands, repeated file writes, and repeated external API calls.

This fix makes restart behavior safer by treating lost workers as a terminal failure for the affected run instead of pretending the run never started. If retries are configured, the system schedules a fresh follow-up attempt explicitly, which preserves retry semantics without silently duplicating the original in-flight execution.

## Validation

- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -229,6 +229,8 @@ const (
 	jobsV2WorkerMinDelay      = 100 * time.Millisecond
 	jobsV2WorkerIdleDelay     = time.Minute
 	jobsV2WorkerErrorDelay    = 200 * time.Millisecond
+
+	jobsV2RecoveryWorkerLostError = "worker lost during crash recovery; run may have partially completed"
 )
 
 type jobsV2ProgramConfig struct {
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as requested; implementing the underlying recovery fix locally.

PR #734, opened 2026-05-28
fix: Web/API sessions never persist user_turns, leaving serve-created session stats permanently wrong

sam-saffron-jarvis:
## What changed

- fixed `cmd/serve_runtime.go` so serve/web/API snapshot persistence now derives the exact `user_turns` count from the snapshot and persists it alongside any summary update
- fixed incremental serve/web/API message appends so persisted user messages also call `IncrementUserTurns(...)`
- kept the runtime session metadata in sync with the persisted `user_turns` value
- added focused tests covering both the snapshot rewrite path and the incremental append path

## Why this is high-value

This was a real persisted-state corruption bug on the serve/web/API path: messages were being saved correctly, but `user_turns` could stay at `0` or otherwise stale forever. That makes session stats wrong for exported data, inspection, and any logic that depends on user-turn counts. The bug specifically affects long-lived web/API sessions, so fixing it improves reliability for an important surface area without changing broader behavior.

## Validation

- `gofmt -w cmd/serve_runtime.go cmd/serve_runtime_test.go`
- `go build ./...`
- `go test ./...`
- added tests:
  - `TestServeRuntimePersistSnapshotUpdatesUserTurns`
  - `TestServeRuntimeAppendMessagesIncrementsUserTurns`
- `git diff --check`

### DIFF
--- a/cmd/serve_runtime.go
+++ b/cmd/serve_runtime.go
@@ -369,6 +369,16 @@ func (rt *serveRuntime) ensurePersistedSession(ctx context.Context, sessionID st
 	return true
 }
 
+func countUserTurns(messages []llm.Message) int {
+	userTurns := 0
+	for _, msg := range messages {
+		if msg.Role == llm.RoleUser {
+			userTurns++
+		}
+	}
+	return userTurns
+}
+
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as requested; implementing the underlying user_turns persistence fix locally.

PR #735, opened 2026-05-28
fix(contain): build runit from void-linux source on arch agent image

arpitjalan:
The Arch agent Dockerfile installed runit via `yay -S runit`, whose AUR PKGBUILD downloads the upstream source tarball from smarden.org. smarden.org is frequently unreachable; `term-llm contain new` stalled for ~17 minutes on step 7/16 of a fresh build before the curl retry loop timed out:

`curl: (28) Failed to connect to smarden.org port 80 after 134517 ms`

Switch to cloning github.com/void-linux/runit and building runsvdir/ runsv/sv/utmpset in place, matching what Dockerfile.fedora already does. Since runit was the only AUR dependency, also drops the `aur` user, sudoers entry, and yay install.

### DIFF
--- a/internal/contain/images/agent/Dockerfile
+++ b/internal/contain/images/agent/Dockerfile
@@ -92,16 +92,27 @@ RUN useradd -m -s /bin/zsh -G wheel agent && \
     chown agent:agent /home/agent/.zshrc && \
     sudo -Hu agent playwright install chromium
 
-# runit from Void Linux's maintained runit tree. Arch's AUR PKGBUILD fetches
-# the upstream tarball from smarden.org, which is frequently unreachable; the
-# Void fork carries modern compiler fixes while preserving runit semantics.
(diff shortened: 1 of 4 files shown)


arpitjalan:
@SamSaffron I hope we're not using the aur user and yay install inside the container. If yes, I'll let the user and yay install be left as-is.

SamSaffron:
yeah yay is deliberate I would keep it, don't mind changing the clone. We could also just create a new AUR package here, also not sure but I think a few runit binaries may be missing there from your cp at the end. 

I worry a bit about stray binaries, it is nice to have it all in the package manager.

arpitjalan:
Reworked it: kept yay, and runit now installs via a local PKGBUILD that points at the Void github fork instead of smarden.org.

SamSaffron:
looks good!

PR #736, opened 2026-05-28
feat(auth): add 'term-llm auth' command for OAuth sign-in

arpitjalan:
ChatGPT and Copilot OAuth previously bootstrapped lazily — the very first `term-llm ask --provider chatgpt ...` would print a device code mid-task and block until the user finished signing in. There was no way to do auth out-of-band, no way to see what's signed in, and no way to sign out without manually deleting the JSON files under ~/.config/term-llm/.

Add a dedicated `auth` command group with three subcommands:

```
term-llm auth login [provider]   # interactive picker if no arg;
                                 # reuses the existing OAuth flow
term-llm auth status             # show signed-in state + expiry
                                 # for every OAuth provider
term-llm auth logout [provider]  # clear stored credentials
```

Scope: ChatGPT (Codex) and GitHub Copilot only. API-key providers (Anthropic, OpenAI direct, Gemini, OpenRouter, XAI, SambaNova, BFL) are intentionally excluded — they're already covered by env vars and `term-llm config set`, so an auth subcommand for them would be a thin "paste your key" wrapper with no UX win.

Implementation:

- Export `PromptForChatGPTAuth` / `PromptForCopilotAuth` so the new command drives the same flow as lazy auth (single source of truth; no duplicated UX, signal handling, or token persistence).
- Lazy auth is unchanged; the non-interactive error messages and the setup-wizard hint now point users at `term-llm auth login` instead of "run an ask interactively first".
- Internal `authProvider` registry keeps the cmd-layer small and makes adding future OAuth providers a single struct entry.
- Status output distinguishes "expired (will refresh on next use)" from "not signed in" — accurate to the actual refresh-token behavior in RefreshChatGPTCredentials.
- Picker uses huh (already a repo dep, same pattern as the setup wizard); stdin-not-a-TTY without an explicit provider returns a clear `pass one of: chatgpt, copilot` error rather than letting huh crash.

### DIFF
--- a/cmd/auth.go
+++ b/cmd/auth.go
@@ -1,0 +1,256 @@
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"charm.land/huh/v2"
+	"github.com/samsaffron/term-llm/internal/credentials"
+	"github.com/samsaffron/term-llm/internal/llm"
+	"github.com/spf13/cobra"
+	"golang.org/x/term"
+)
+
+// authProvider describes an OAuth provider we can drive a sign-in flow for.
+// Each provider exposes the four hooks the auth subcommands need; we keep
+// the registry small (just ChatGPT + Copilot) because API-key providers
+// are configured via env vars or `term-llm config set`.
+type authProvider struct {
+	name     string // user-facing label
+	id       string // CLI slug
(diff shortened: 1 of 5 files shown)


SamSaffron:
thanks, yeah seems useful

PR #743, opened 2026-05-29
feat(serve): hand long chat work to jobs

sam-saffron-jarvis:
## What

Add a first-class chat → jobs handoff path for long-running work.

This introduces a new server tool:

- `handoff_to_job`

When the web runtime is started with jobs enabled (`term-llm serve web jobs`), an agent can call this tool to:

1. create a manual LLM job
2. trigger it immediately
3. return `{job_id, run_id, session_id}` to the current chat
4. have the job append its final result back to the originating chat session when it finishes

## Why

Some foreground chat runs legitimately become long tool-heavy loops. The right fix is not just "raise the timeout"; it is to let the live chat hand control to a durable job and return a handle immediately.

This is deliberately minimal:

- no new UI surface
- no new scheduler model
- uses the existing jobs runner
- uses normal session persistence for the completion reply
- bounded default timeout: 2h, capped at 6h for handoff requests

## Notes

The tool is only available when jobs are present in the same serve process. That keeps the first version small and avoids inventing cross-process job RPC/config.

## Validation

```text
go test ./cmd ./internal/tools
go build ./...
```

### DIFF
--- a/cmd/job_handoff_test.go
+++ b/cmd/job_handoff_test.go
@@ -3,6 +3,8 @@ package cmd
 import (
 	"context"
 	"encoding/json"
+	"net/http"
+	"net/http/httptest"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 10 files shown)


SamSaffron:
at last I did something here

PR #744, opened 2026-05-29
fix: Responses continuations do a full session DB scan before checking in-memory history

sam-saffron-jarvis:
## What changed

- Updated `populateResponsesToolResultNames` to consult `runtime.snapshotHistory()` before touching the session store.
- Only fall back to `s.store.GetMessages(...)` when some tool-call IDs are still unresolved after checking in-memory history.
- When the store fallback is needed, scan persisted messages from newest to oldest and stop as soon as every missing tool result name has been found.
- Removed the extra allocation/conversion step that previously rebuilt every persisted row as `llm.Message` before scanning it.
- Added tests covering both the fast path (runtime history resolves names without any DB read) and the fallback path (store still fills unresolved names).

## Why this is high-value

Response continuations call this helper on the request path before streaming starts. Previously, every continuation with unnamed tool results did a full `GetMessages` read for the session and converted every stored row to `llm.Message` before even checking whether the active runtime already had the needed tool-call names in memory.

This change reduces first-token latency and total request work for long-lived serve/web sessions because the common case can now be satisfied entirely from in-memory history, avoiding:

- a full SQLite session-history read
- JSON/row decoding for all persisted messages
- allocation of a second `[]llm.Message` copy of the whole stored history

Even on fallback, the code now stops scanning once all requested tool-call names are found instead of always converting and traversing the entire persisted history.

## Validation

- `gofmt -w cmd/serve_handlers_responses.go cmd/serve_test.go cmd/serve_runtime_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --check`

Added tests:
- `TestPopulateResponsesToolResultNames_UsesRuntimeHistoryBeforeStore`
- `TestPopulateResponsesToolResultNames_FallsBackToStoreForUnresolvedNames`

### DIFF
--- a/cmd/serve_handlers_responses.go
+++ b/cmd/serve_handlers_responses.go
@@ -402,37 +402,41 @@ func (s *serveServer) populateResponsesToolResultNames(ctx context.Context, sess
 	}
 
 	names := map[string]string{}
+	collectParts := func(parts []llm.Part) {
+		for _, part := range parts {
+			if part.Type != llm.PartToolCall || part.ToolCall == nil {
+				continue
+			}
+			id := strings.TrimSpace(part.ToolCall.ID)
+			if id == "" || !missing[id] || names[id] != "" {
+				continue
+			}
+			if name := strings.TrimSpace(part.ToolCall.Name); name != "" {
+				names[id] = name
+			}
+		}
+	}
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #745, opened 2026-05-29
fix: Config saves are non-atomic and can corrupt config.yaml on interruption

sam-saffron-jarvis:
## What changed

- replaced the direct `os.WriteFile` calls used by config saves in `internal/config/config.go` with a small `writeFileAtomically` helper
- the helper writes to a temp file in the same directory, `fsync`s it, applies the expected mode, and then atomically renames it into place
- updated both config write paths covered by this issue:
  - `rewriteProviderEnvSections`
  - `Save`
- added focused tests that verify:
  - atomic writes replace the file contents and enforce `0600`
  - if temp-file creation fails, the existing `config.yaml` contents remain untouched

## Why this is high-value

`~/.config/term-llm/config.yaml` is core persisted state for provider defaults and auth-related settings. Before this change, these saves truncated the live file and then rewrote it in place. If the process was interrupted mid-write, or the write only partially completed, users could be left with an empty or malformed config and broken startup/auth flows.

With the temp-file + sync + rename pattern, the old config stays intact until the replacement file is fully written and ready, which removes the torn-write corruption mode from these paths.

## Validation

- `gofmt -w internal/config/config.go internal/config/config_test.go`
- `go build ./...`
- `go test ./internal/config`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -940,7 +940,44 @@ func rewriteProviderEnvSections(path string, snapshot map[string]map[string]stri
 		return err
 	}
 	enc.Close()
-	return os.WriteFile(path, buf.Bytes(), 0600)
+	return writeFileAtomically(path, buf.Bytes(), 0o600)
+}
+
+func writeFileAtomically(path string, data []byte, perm os.FileMode) error {
+	dir := filepath.Dir(path)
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #746, opened 2026-05-29
fix: Telegram restart carryover loads and converts the entire prior session just to keep a small tail

sam-saffron-jarvis:
## What changed

- Reworked Telegram carryover restore to page backward from the end of the prior session instead of calling `GetMessages(..., 0, 0)` and materializing the full transcript.
- Added a reverse-order session-store read path (`GetMessagesPageDescending`) on the SQLite store, plus `LoggingStore` forwarding/fallback.
- Stopped converting all persisted rows to `llm.Message`; carryover restore now only converts the suffix needed to satisfy `TelegramCarryoverChars`.
- Added regression tests for:
  - Telegram carryover restoring the right tail without reading the whole transcript
  - Reverse descending pagination in the SQLite session store

## Why this is high-value

Telegram restart recovery runs on the first message after a bot restart/new session handoff, so this path is user-visible and on the critical recovery path.

Before this change, `restoreHistoryFromDB`:

- loaded the entire prior session from SQLite,
- deserialized every message row,
- converted every row to `llm.Message`,
- and only then trimmed to a small carryover tail.

That means long-running Telegram chats paid full-history read/decode/allocation cost even when only a tiny suffix was kept. The new path reduces work and memory pressure by:

- reading only reverse pages from the tail,
- stopping as soon as the carryover character budget is satisfied,
- and converting only the kept suffix (plus at most one overshoot message).

This directly reduces restart latency, unnecessary allocations, GC pressure, and OOM risk for large Telegram histories.

## Validation

- `gofmt -w internal/serve/telegram.go internal/serve/telegram_test.go internal/session/sqlite.go internal/session/sqlite_test.go internal/session/logger.go`
- `go build ./...`
- `go test ./...`
- Added/ran targeted regression tests covering the new paged carryover path and SQLite descending pagination.

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -578,26 +578,93 @@ func (m *telegramSessionMgr) restoreHistoryFromDB(ctx context.Context, chatID in
 			continue
 		}
 
-		msgs, loadErr := m.store.GetMessages(ctx, s.ID, 0, 0)
+		history, loadedRows, loadErr := m.loadCarryoverHistory(ctx, s.ID, maxChars)
 		if loadErr != nil {
 			log.Printf("[telegram] restore history: get messages failed for session %s: %v", s.ID, loadErr)
 			continue
 		}
 
-		// Convert all messages, then take only the tail that fits within maxChars.
+		history = sanitizeCarryoverMessages(history)
+		if len(history) > 0 {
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #747, opened 2026-05-29
feat(contain): add 'reauth' to seed ChatGPT OAuth into a running workspace

arpitjalan:
Workspaces created with `term-llm contain new` could end up unauthenticated for ChatGPT even when
the host was signed in (later). The container's web UI reported "not signed in" while `term-llm auth status` on the host
showed a fresh token, and the only recovery was `term-llm contain rm` + recreate — which destroys the
workspace's named volume and all its state.

The common path into this state: signing in to ChatGPT on the host *after* creating the workspace. `contain new`
reads ~/.config/term-llm/chatgpt_oauth.json once at create time and bakes its base64 into the rendered .env
(TERM_LLM_CHATGPT_OAUTH_JSON_B64). If the file doesn't exist yet, the slot is rendered empty. The container then
boots, writes a sentinel that blocks any future hydration from .env, and ends up with no credentials. The host
login that happens later never gets copied in — `contain new` already again. And because the empty-slot render
is silent, the user has no in-CLI signal that this trap was set. The same trap also fires when the user picks a
non-chatgpt provider initially or never signs in on the host at all.

Add `term-llm contain reauth <name>`: reads ~/.config/term-llm/chatgpt_oauth.json on the host and writes it to
<workspace>/.config/term-llm/chatgpt_oauth.json inside the container via `docker compose exec -T --user
<label-user> <service> sh -c 'mkdir -p ...; umask 077; cat > ...; chmod 600 ...'`. The container's term-llm
picks it up on the next call; long-running processes (serve, jobs) need a restart, which the command prints as a
follow-up hint.

Also surface the recovery path at the moment the trap is set: `contain new` now ends its "Next steps:" block
with one line when the rendered .env has TERM_LLM_PROVIDER=chatgpt and TERM_LLM_CHATGPT_OAUTH_JSON_B64= empty:

```
  ChatGPT auth not seeded; after `term-llm auth login chatgpt` on
  the host run: term-llm contain reauth <name>
```

### DIFF
--- a/cmd/contain.go
+++ b/cmd/contain.go
@@ -128,6 +128,26 @@ This is destructive and cannot be undone. You must type YES to continue.`,
 	},
 }
 
+var containReauthCmd = &cobra.Command{
+	Use:   "reauth <name>",
+	Short: "Copy the host's ChatGPT OAuth credentials into a running contain workspace",
+	Long: `Copy the host's chatgpt_oauth.json into a running contain workspace.
+
+This reads ~/.config/term-llm/chatgpt_oauth.json on the host and writes it to
+<workspace>/.config/term-llm/chatgpt_oauth.json inside the container via
+docker compose exec. Useful when 'term-llm contain new' ran before the host
+was signed in, or when you want to share a fresh host login with an existing
+workspace without rebuilding the volume.
+
+The workspace must be running. Long-running processes inside the container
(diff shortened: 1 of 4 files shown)


SamSaffron:
I think I am preferring: 

2ccb82d56ed4179509d79aa007d7910378a03e7c

It is more flexible anyway

PR #748, opened 2026-05-30
fix: spawn_agent drops partial subagent output and child session ID on timeout/error

sam-saffron-jarvis:
## What changed

- Updated `SpawnAgentTool.Execute` to preserve `runResult.Output` and `runResult.SessionID` when the subagent runner returns an error.
- Added a small helper so timeout, cancellation, max-turns, and generic execution-failure responses keep the existing structured error fields while also carrying any partial subagent result.
- Added tests covering both generic failures and timeout-classified failures with partial output/session IDs.

## Why this is high-value

`cmd/spawn_runner.go` already returns partial subagent output plus the child session ID on failure. Before this change, `spawn_agent` threw that away in every error branch and returned only synthesized error JSON. That meant a timed-out or failing delegated investigation could do useful work, create a child session, and still lose all user-visible output and the session link. This fix prevents that data loss without changing the success path or broadening behavior elsewhere.

## Validation

- `gofmt -w internal/tools/spawn_agent.go internal/tools/spawn_agent_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/tools/spawn_agent.go
+++ b/internal/tools/spawn_agent.go
@@ -311,18 +311,18 @@ func (t *SpawnAgentTool) Execute(ctx context.Context, args json.RawMessage) (llm
 
 	if err != nil {
 		if llm.IsMaxTurnsExceeded(err) {
-			return llm.TextOutput(t.formatErrorWithDuration(ErrExecutionFailed, fmt.Sprintf("agent '%s' stopped after reaching max turns: %v", a.AgentName, err), duration)), nil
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented in the working branch: spawn_agent now preserves partial subagent output/session IDs on timeout and error paths, with regression tests. Verified with go build ./... and go test ./....

PR #749, opened 2026-05-30
fix: grep raw-output safety cap before full-line allocation

sam-saffron-jarvis:
## What changed

- replaced `runRipgrep`'s raw stdout capture loop with a chunked `captureRipgrepOutput` helper built on `bufio.Reader.ReadSlice('\n')`
- enforce `rgMaxBufferedBytes` before assembling an entire ripgrep line in memory
- keep only complete lines in the buffered output so truncated captures still hand valid line-delimited data to the existing parser
- added tests covering:
  - stopping reads near the byte cap for a single oversized line
  - preserving completed lines while dropping the oversized partial line that triggered truncation

## Why this is high-value

`grep` is on a hot path for code review and repository exploration. The previous implementation used `ReadBytes('\n')`, which allocates the full line before checking the 8 MB cap. With ripgrep `--json`, one generated/minified source line can become one enormous JSON record, so a single match could force large transient allocations and extra read latency exactly where the code is supposed to stay bounded and responsive.

This change makes the cap real:

- memory is bounded by `out + current line fragment buffer`, instead of an unbounded whole-line allocation
- oversized lines are cut off after roughly one reader-buffer worth of extra input, instead of being fully read first
- truncated output still preserves prior complete records, so behavior stays aligned with the existing parser and formatting pipeline

## Validation

- `go build ./...`
- `go test ./internal/tools`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/tools/grep.go
+++ b/internal/tools/grep.go
@@ -176,6 +176,67 @@ func (l ripgrepCaptureLimits) normalize() ripgrepCaptureLimits {
 	return l
 }
 
+func captureRipgrepOutput(stdout io.Reader, kill func() error, limits ripgrepCaptureLimits) ([]byte, bool, error) {
+	limits = limits.normalize()
+
+	var out bytes.Buffer
+	var lineBuf bytes.Buffer
+	reader := bufio.NewReader(stdout)
+	lineCount := 0
+	truncated := false
+
+	flushLine := func() {
+		lineCount++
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented in the working branch: grep raw capture now enforces the byte cap before assembling oversized lines and preserves complete prior lines, with regression tests. Verified with go build ./... and go test ./....

PR #750, opened 2026-05-30
fix: config set rewrites config.yaml non-atomically and can corrupt the main config on interruption

sam-saffron-jarvis:
## What changed

- Replaced `term-llm config set`'s direct `os.WriteFile` rewrite with the existing temp-file + `fsync` + `rename` atomic write path from `internal/config`.
- Preserved the existing config file's mode when replacing an existing `config.yaml`, while still defaulting to `0644` for first-time creation.
- Added a regression test that makes atomic temp-file creation fail and verifies the existing `config.yaml` stays unchanged.

## Why this is high-value

`config.yaml` is shared persisted state for the CLI, serve mode, jobs, Telegram, and provider credentials. Before this change, `config set` truncated and rewrote the file in place, so an interruption during the write could leave the main config empty or partially written and break future startups until the file was repaired manually.

This change prevents that failure mode by only swapping the new file into place after the full contents have been written and synced successfully.

## Validation

- `gofmt -w cmd/config.go cmd/config_test.go internal/config/config.go`
- `go build ./...`
- `go test ./...`
- Added `TestConfigSet_AtomicWriteFailureLeavesExistingConfigUntouched`
- `git diff --check`

### DIFF
--- a/cmd/config.go
+++ b/cmd/config.go
@@ -1019,6 +1019,13 @@ func configSet(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to get config path: %w", err)
 	}
 
+	writePerm := os.FileMode(0o644)
+	if info, err := os.Stat(configPath); err == nil {
+		writePerm = info.Mode().Perm()
+	} else if !os.IsNotExist(err) {
+		return fmt.Errorf("failed to stat config: %w", err)
+	}
+
 	// Ensure config directory exists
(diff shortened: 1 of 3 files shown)


SamSaffron:
Implemented in the working branch: config set now writes through the atomic temp-file/fsync/rename path while preserving existing file mode, with regression tests. Verified with go build ./... and go test ./....

PR #751, opened 2026-05-31
feat(jobs): bring native llm runner to parity with ask (model, cwd, tools, files)

sam-saffron-jarvis:
> ### ⚠️ Revised after review (commit `101a2821`)
>
> Two changes from the original proposal below:
>
> 1. **`cwd` is now REQUIRED** for `llm` jobs (validated in `jobsV2LLMRunner.Run`
>    before execution). An `llm` job must declare where its file/shell tools are
>    rooted rather than silently inheriting the jobs server's process working
>    directory. The field loses `omitempty`. The `program` runner's optional
>    `cwd` is unchanged (it is an explicit escape hatch).
> 2. **The `files` toggle was dropped.** It only folded file contents into the
>    user prompt (pure `ask -f` parity, no new capability); `instructions`
>    already covers that. The now-unused `input`/`prompt` imports were removed.
>
> New test `TestJobsV2LLMRunnerRequiresCwd` covers the rejection (omitted + blank
> `cwd`, executor not invoked). References to `files` and the optional-`cwd`
> behavior in the description below are superseded by this note.
>
> ---

## What

Bring the jobs v2 **native `llm` runner** to parity with the `ask` CLI by adding
optional `runner_config` toggles, while keeping the `program` runner as the
escape hatch. Every new field is optional and `omitempty`; omitting it preserves
today's exact behavior.

New `jobsV2LLMConfig` fields (`cmd/serve_jobs_v2.go`):

| field | meaning |
|-------|---------|
| `provider` | provider override, e.g. `"chatgpt"` or `"chatgpt:gpt-5.5-xhigh"` |
| `model` | model override (alternative to `provider:model`) |
| `cwd` | per-run tool root (file roots + shell `exec.Cmd.Dir`) — **never `os.Chdir`** |
| `read_dir` / `write_dir` | additional tool roots |
| `tools` | tool-set override (`"all"` or csv) |
| `max_turns` | agentic turn cap |
| `max_output_tokens` | output-token cap |
| `files` | context files attached to the prompt (like `ask -f`) |
| `search` | enable web search |
| `system_message` | system-prompt override |
| `skills` | skills selection |

Threaded onto the **same execution path `ask` already uses** via two small,
independently testable helpers (`applyJobLLMProviderModel`,
`resolveJobLLMSettings`) reusing the existing override machinery
(`applyProviderOverridesWithAgent` / `applyAgentModelOverride`).

## Why

Today the native `llm` runner can't set a model or a working directory, so the
only way to run (say) two `developer` jobs on different models in different git
worktrees is to drop to a `program` job shelling out to `term-llm ask`. That
escape hatch throws away everything that makes the native runner valuable:
session linkage (`session_id` on the run), progressive best-so-far state, and the
structured event timeline.

**Decision: keep both runners; give the native one parity.** `program` stays the
"run an arbitrary thing" escape hatch; `llm` becomes a first-class native run
that no longer forces a subprocess just to set a model and a cwd.

## The `cwd` constraint: no `os.Chdir`

The jobs server runs many runs concurrently in **one process**, so a global
`os.Chdir` would race every other run. `cwd` is implemented as "root this run's
file/shell tools at this directory":

- `cwd` is appended to the run's read/write tool roots, and
- a new `ToolConfig.ShellWorkingDir` (via `SessionSettings.ShellWorkingDir`)
  becomes the shell tool's default `exec.Cmd.Dir` when a call omits `working_dir`
  (an explicit `working_dir` still wins).

No `os.Chdir` anywhere; a test captures `os.Getwd()` before/after a full run and
asserts it is unchanged.

**Scope note (v1):** relative paths passed to the *file* tools still resolve
against the process cwd (the file tools have no per-call base dir, and a process
chdir is forbidden here). The shell — the common case for the worktree scenario
(build/test/git) — is correctly rooted at `cwd`. Rerooting relative file-tool
paths without a global chdir is a follow-up.

## Tests

New table-driven tests (`cmd/serve_jobs_v2_parity_test.go`, +`internal/tools/shell_test.go`):
config round-trip + `omitempty`; provider/`provider:model`/model override with
**deep-copy isolation** asserted (no mutation of shared config); **cwd routing
with `os.Getwd()` asserted unchanged**; shell `Dir == cwd` (explicit `working_dir`
wins); `max_turns`/`max_output_tokens`/`tools`/`read_dir`/`write_dir`/`search`/
`system_message`/`files` reach exec opts with serve-flag fallbacks; skills
threading; and a focused integration test driving the **real** serve-jobs
executor (hermetic `debug` provider) that creates its output rooted in `cwd`.

Gates: `gofmt -l` clean · `go build ./...` · `go vet ./...` · `go test ./...` — all green.

## Backward compatibility

All new fields optional; existing job definitions deserialize and resolve to
identical settings. No change to the `program` runner or the run/event schemas
beyond additive config.

## Reviewer note

One commit (`chore: fix pre-existing go vet warnings`) is **out of this feature's
scope** — it fixes two pre-existing `lostcancel` leaks in `cmd/progressive.go`
(behavior-preserving) and three test-file vet warnings in `internal/llm`, needed
to make `go vet ./...` green. It's isolated in its own commit; happy to split it
into a separate PR if you'd prefer.

---

🤖 Generated with term-llm (developer agent, `claude-bin:opus-max`) from design
doc `job-runner-parity.md`.


SamSaffron:
Closing for follow-up cleanup: required cwd is a breaking behavior change, docs/examples need updating, and validation should happen at job create/update rather than only at run time.

PR #752, opened 2026-05-31
fix: cancel standalone web interrupts on fallback

sam-saffron-jarvis:
## What

Fix interrupt classification fallback so standalone replacement prompts sent while a web stream is active cancel the stale run instead of being injected into it.

Tiny repros I used:

- Session #2378 had the bad shape: Sam asked `kanazawa vs kyoto`, but the active run kept chasing the old Okinawa/Places thread.
- Live claude-bin recall chain works when continuation is clean: `231` → `234` → `8733, what previous?` returned `231, 234`.
- Added targeted tests for the failure mode: `kanazawa vs kyoto` during an active Okinawa task now classifies as `cancel`, while steering prompts like `also include quiet islands`, `what about non beach places`, and `10 non beach places as well` remain `interject`.

## Why

The web UI sends messages during an active stream through the interrupt endpoint. If the fast classifier is unavailable or times out, the old fallback always returned `interject`. That is safe for corrections, but terrible for standalone new prompts: they get queued into the stale run and the assistant keeps answering the old task. Technically working. Socially deranged.

## Verification

```text
go test ./internal/llm -run 'TestClassifyInterrupt|TestClassify$|TestClassifyTimeout'
go test ./cmd ./internal/llm
go build ./...
```

### DIFF
--- a/internal/llm/classify.go
+++ b/internal/llm/classify.go
@@ -90,6 +90,50 @@ func classifyInterruptHeuristic(msg string) (InterruptAction, bool) {
 	return InterruptInterject, false
 }
 
+func classifyInterruptFallback(msg string, activity InterruptActivity) InterruptAction {
+	normalized := strings.ToLower(strings.TrimSpace(msg))
+	if normalized == "" {
+		return InterruptInterject
+	}
+
+	// If the user is obviously amending the current work, keep the current run
+	// alive and inject the note at the next legal boundary.
+	steeringPrefixes := []string{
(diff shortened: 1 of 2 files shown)


SamSaffron:
Porting this classifier fallback fix directly into the working branch instead of merging this PR. The change (classifyInterruptFallback + the two fallback call sites in ClassifyInterrupt, plus the standalone-cancel / steering-interject tests) has been applied and verified there: go test ./internal/llm and ./cmd pass, go build ./... is clean. Closing as superseded.

PR #754, opened 2026-05-31
feat(contain): auto-select default web port and reconcile config on start

arpitjalan:
`contain new` previously defaulted the Web UI port prompt to a hard-coded 8081, so every workspace collided unless the port was typed manually. Add a `generate: web_port` template default that scans existing workspace .env files and probes host port availability, returning the lowest free port at or above 8081. Resolution runs before the workspace directory is created, so it never collides with itself.

The port value was also silently ignored on restart: Start() ran `docker compose start` when a container already existed, which boots the container with its baked-in config and never applies drift from an edited .env/compose.yaml. A workspace whose port was changed kept its stale binding. Switch Start() to `docker compose up -d` in both the create and existing- container paths so config drift is reconciled; managed image build/sync still only runs when creating from scratch. `up -d` is idempotent on an unchanged running container (verified: no recreate, no downtime), and recreates only on genuine drift, where volume-backed state (/home/agent) is preserved.

### DIFF
--- a/internal/contain/docker.go
+++ b/internal/contain/docker.go
@@ -85,14 +85,18 @@ func Start(ctx context.Context, runner Runner, name string, stdout, stderr io.Wr
 	if err := ensureComposeDefinition(name); err != nil {
 		return err
 	}
-	if hasContainers, err := composeHasContainers(ctx, runner, args, dir); err != nil {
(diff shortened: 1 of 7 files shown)


SamSaffron:
Thanks Arpit

SamSaffron:
going to do a little followup on this as well.

PR #756, opened 2026-05-31
fix: grep max_results is applied per file, so common searches still do huge rg scans and buffering

sam-saffron-jarvis:
## What changed

- kept ripgrep's existing `--max-count` behavior as a per-file guard, but stopped treating it as the global `max_results` limiter
- added a streaming stdout consumer in `internal/tools/grep.go` so JSON ripgrep output is inspected line-by-line while it is being captured
- added a small `ripgrepResultLimiter` that mirrors the existing `parseToPending` flush boundaries and tells `runRipgrep` when the requested total match count has already been satisfied
- when that global cap is reached, `runRipgrep` now kills `rg` immediately instead of continuing to scan, emit JSON, buffer output, and only truncate later
- added a regression test that feeds a large synthetic ripgrep stream and verifies capture stops after the 100th global result while still leaving a parseable useful prefix

## Why this is high-value

Today `max_results` only limits ripgrep matches per file, while `term-llm` still captures the full raw `rg` stream before parsing. For common patterns across many files, searches asking for a small number of results can still do a large amount of unnecessary filesystem scanning, JSON serialization, pipe I/O, and memory allocation.

This fix is high-value because it moves the stop condition into the streaming capture path. Once enough total matches are found, we terminate `rg` immediately. That directly reduces wasted work, lowers allocation and buffering pressure, and improves interactive latency for code-review and exploration flows.

## Validation

- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/tools/grep.go
+++ b/internal/tools/grep.go
@@ -132,6 +132,8 @@ func buildRipgrepArgs(a GrepArgs, searchPath string, contextLines, maxResults in
 	} else {
 		args = append(args,
 			"--json",
+			// --max-count is per-file, not global. A streaming stop enforces the
+			// caller's overall max_results limit once enough total matches are found.
 			"--max-count", strconv.Itoa(maxResults),
 			"--context", strconv.Itoa(contextLines),
 		)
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented the important changes from this PR directly in main worktree.

PR #757, opened 2026-06-01
fix: Claude Bin stream can deadlock forever when the event consumer stops reading

sam-saffron-jarvis:
## What changed

- close the Claude stdout scanner bridge as soon as `dispatchClaudeEvents(...)` returns an error, instead of waiting for the deferred cleanup at function exit
- guard `bridge.done` with `sync.Once` so the early unblock and deferred cleanup can safely share the same close path
- add a regression test that uses a fake `claude` binary to verify `runClaudeCommand` returns instead of hanging when the event consumer has stopped reading and Claude keeps streaming output

## Why this is high-value

A downstream streaming write failure could previously wedge Claude-backed requests forever. The failure mode was:

1. event delivery failed and `dispatchClaudeEvents(...)` returned early
2. the stdout scanner goroutine kept trying to enqueue parsed lines into `lineCh`
3. once `lineCh` filled, the scanner blocked forever because `bridge.done` was still only scheduled to close in deferred cleanup
4. `runClaudeCommand` then waited forever on `scanErrCh`

That can strand in-flight Claude executions in serve/TUI/Telegram flows after client disconnects or other stream write failures. Unblocking the scanner immediately on the error path prevents the deadlock.

## Validation

- `gofmt -w internal/llm/claude_bin.go internal/llm/claude_bin_test.go`
- `go build ./...`
- `go test ./...`
- added regression test: `TestRunClaudeCommand_UnblocksScannerWhenEventConsumerStopsReading`

### DIFF
--- a/internal/llm/claude_bin.go
+++ b/internal/llm/claude_bin.go
@@ -776,6 +776,12 @@ func (p *ClaudeBinProvider) runClaudeCommand(
 		toolReqCh: make(chan claudeToolRequest, 64),
 		done:      make(chan struct{}),
 	}
+	var bridgeDoneOnce sync.Once
+	stopScanner := func() {
+		bridgeDoneOnce.Do(func() {
+			close(bridge.done)
+		})
+	}
 	p.eventsMu.Lock()
 	p.currentBridge = bridge
 	p.currentEvents = send.ch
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #758, opened 2026-06-01
fix: Claude CLI subprocesses are launched without process-group cleanup or WaitDelay

sam-saffron-jarvis:
## What changed

- added a small `prepareClaudeCommand` helper in `internal/llm/claude_bin.go`
- configured Claude CLI subprocesses with the same hygiene used elsewhere in the repo:
  - `procutil.PrepareCommand(cmd)` so the command runs in its own process group and context cancellation targets the whole group
  - `cmd.WaitDelay = time.Second` to bound teardown when descendants keep stdio pipes open
- changed the dispatch-error cleanup path to call `cmd.Cancel()` instead of killing only the direct Claude pid
- added a unit test that verifies Claude command setup includes process-group isolation, process-group cancellation, and `WaitDelay`

## Why this is high-value

Claude-backed sessions are a core interactive path for TUI, serve, and Telegram. Before this change, a cancelled or failed Claude turn only killed the immediate `claude` process. Any helper or MCP-related descendants could survive cancellation, keep stdout/stderr pipe fds open, and leave behind orphaned processes or stuck teardown. This fix makes cancellation hit the whole subprocess tree and adds the same bounded-wait behavior already used in other long-running runners in the codebase.

## Validation

- `gofmt -w internal/llm/claude_bin.go internal/llm/claude_bin_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/llm/claude_bin.go
+++ b/internal/llm/claude_bin.go
@@ -9,6 +9,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"log/slog"
 	"os"
 	"os/exec"
@@ -19,6 +20,7 @@ import (
 	"time"
 
 	"github.com/samsaffron/term-llm/internal/mcphttp"
+	"github.com/samsaffron/term-llm/internal/procutil"
 )
 
 // claudeStderrTailMaxLines caps the number of trailing stderr lines we retain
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #759, opened 2026-06-01
fix: Serve runtime reserializes and rescans the full transcript on every persisted turn

sam-saffron-jarvis:
## What changed

- Added a `historyPersisted` runtime flag so serve/web sessions can remember when the in-memory active transcript already matches the persisted session state.
- Switched the incremental persistence gate from `HasCompactionBoundary(...)` to that trusted persisted-history flag, so later persisted turns stay on the append-only path instead of reseeding with `ReplaceMessages(...)` every turn.
- Kept conservative fallbacks: the runtime now drops back to a full snapshot if append-only writes miss part of the turn, if an assistant row update/add fails, or if history was mutated outside the normal persistence flow.
- Added focused tests for:
  - two consecutive persisted turns only doing one initial `ReplaceMessages(...)`
  - append-only turns still reconciling with a full snapshot after an injected assistant-update failure

## Why this is high-value

Before this change, non-compacted persisted serve sessions re-ran an O(history size) snapshot replace on the first callback of every turn, which forces full transcript reserialization and a SQLite prefix scan even when only the new suffix changed. That cost grows with conversation length and directly affects the main serve/web path.

After this change, once a session history is known-good in persistence, later turns append only the new input/output rows. That avoids repeated whole-history JSON work and DB reads on long-running sessions while preserving the existing full-snapshot recovery path when persistence falls behind.

## Validation

- `gofmt -w cmd/serve_runtime.go cmd/serve_runtime_test.go cmd/serve_model_swap.go`
- `go build ./...`
- `go test ./cmd -run 'TestServeRuntime(LaterTurnsStayAppendOnlyOnceHistoryPersisted|AppendOnlyTurnFallsBackToSnapshotAfterAssistantUpdateFailure|SuccessfulRunSkipsFinalSnapshotRewrite|FinalSnapshotReconcilesEarlierAssistantUpdateFailure)'`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/cmd/serve_model_swap.go
+++ b/cmd/serve_model_swap.go
@@ -230,6 +230,7 @@ func seedRuntimeHistory(rt *serveRuntime, history []llm.Message) {
 	rt.mu.Lock()
 	defer rt.mu.Unlock()
 	rt.history = copyLLMMessageSlice(history)
+	rt.historyPersisted = false
 	rt.cumulativeUsage = llm.Usage{}
 	rt.lastInjectedPlatform = ""
 	if rt.engine != nil {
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #762, opened 2026-06-02
fix: Every shell/custom-tool run scans all of /proc to reap descendants, adding avoidable latency to the hottest tool path

sam-saffron-jarvis:
## What changed

- Added a `sweepTaggedDescendantsOnSuccess` policy knob to `prepareToolCommand`.
- The shell tool now only keeps the old always-scan behavior for commands that actually use shell operators (`&`, pipes, redirects, subshells, etc.).
- Simple foreground shell commands like `pwd`, `ls`, and `git status` now skip the nonce-based `/proc` walk on normal success.
- Timeouts/cancellations still force the nonce-based reap, and custom tools / run-agent scripts keep the existing always-sweep behavior.
- Added tests that pin down both sides of the behavior:
  - a foreground command no longer triggers the tagged descendant sweep
  - a detached `setsid` child still gets reaped when the caller opts into the always-sweep path

## Why this is high-value

The shell tool is one of the hottest paths in term-llm. Before this change, every successful shell invocation paid for a full `/proc` scan and `/proc/<pid>/environ` reads just to confirm there was nothing to reap.

That work is O(number of processes on the host), adds avoidable I/O and cache churn, and can dominate trivial commands. Skipping that scan for normal foreground shell commands removes the expensive teardown work from the common case while preserving the stricter cleanup path for cancellations and shell commands that are actually likely to create background descendants.

## Validation

- `gofmt -w internal/tools/exec_helpers.go internal/tools/exec_helpers_test.go internal/tools/shell.go internal/tools/custom_tool.go internal/tools/run_agent_script.go`
- `go build ./...`
- `go test ./...`
- Added focused tests in `internal/tools/exec_helpers_test.go` covering:
  - no tagged sweep for a simple foreground command
  - tagged sweep still occurring for a detached `setsid` child when always-sweep mode is enabled

### DIFF
--- a/internal/tools/custom_tool.go
+++ b/internal/tools/custom_tool.go
@@ -119,7 +119,7 @@ func (t *CustomScriptTool) Execute(ctx context.Context, args json.RawMessage) (l
 	}
 	cmd.Env = env
 
-	cleanup, prepErr := prepareToolCommand(cmd)
+	cleanup, prepErr := prepareToolCommand(cmd, true)
 	if prepErr != nil {
 		return llm.TextOutput(formatToolError(NewToolErrorf(ErrExecutionFailed, "script setup error: %v", prepErr))), nil
 	}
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #766, opened 2026-06-03
fix: session Update clobbers user_turns from stale in-memory structs

sam-saffron-jarvis:
## What changed

- Removed `user_turns` from `SQLiteStore.Update` so metadata-only updates no longer write a caller-supplied turn count back into `sessions`.
- Updated the method comment to treat `user_turns` like the other counters that must be mutated through atomic store paths.
- Added `TestUpdateDoesNotClobberUserTurns` to cover the failure mode where `IncrementUserTurns` is followed by `Update` on a stale in-memory session struct.

## Why this is high-value

`user_turns` is a durable counter used for user-visible state and downstream logic. In the current code, `IncrementUserTurns` atomically increments the database value, but a later `Update` call can overwrite that increment with a stale in-memory value. The Telegram path already does exactly that when it increments the counter and then persists the first-message summary without first updating `sess.meta.UserTurns`. This fix removes the clobber path entirely and keeps turn-count mutations on the atomic update path.

## Validation

- Verified the issue is real in current code by inspecting `internal/session/sqlite.go` and the Telegram call path in `internal/serve/telegram.go`.
- Added and ran targeted tests:
  - `go test ./internal/session -run 'TestUpdateDoesNotClobber(TokenMetrics|UserTurns)$'`
- Ran full validation:
  - `go build ./...`
  - `go test ./...`
- Ran diff hygiene checks:
  - `git diff --check`

### DIFF
--- a/internal/session/sqlite.go
+++ b/internal/session/sqlite.go
@@ -1159,9 +1159,9 @@ func (s *SQLiteStore) GetByPrefix(ctx context.Context, prefix string) (*Session,
 
 // Update modifies an existing session's metadata fields.
 // Token metrics (input_tokens, cached_input_tokens, cache_write_tokens, output_tokens)
-// and turn counters (llm_turns, tool_calls) are intentionally excluded — they are
-// managed exclusively by UpdateMetrics (which uses atomic increments) to prevent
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #767, opened 2026-06-04
fix: Telegram stream watchdog can leak a blocked Recv goroutine on timeout/error paths

sam-saffron-jarvis:
## What changed

- Added a `sendStreamDone` helper guarded by `sync.Once` so the Telegram stream watchdog and stream consumer can only publish one terminal result to `streamDone`.
- Switched the watchdog timeout path, `Recv` EOF/error path, and `EventError` path to use that single-completion helper.
- Added a focused regression test that verifies a second completion attempt returns immediately instead of blocking behind the first buffered send.

## Why this is high-value

`streamReply` had multiple concurrent producers writing to a single buffered `streamDone` channel. If one producer filled the channel and `streamReply` returned on another path before draining it, a later blocking send from the stream consumer goroutine could hang forever. In a long-running Telegram bot, repeated stalled or timing-out streams could accumulate leaked goroutines and retained stream state.

This fix is small, low-risk, and directly removes that class of goroutine leak without changing the surrounding stream behavior.

## Validation

- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -1140,6 +1140,12 @@ func (m *telegramSessionMgr) handleMessage(ctx context.Context, bot *tgbotapi.Bo
 	}
 }
 
+func sendStreamDone(done chan<- error, once *sync.Once, err error) {
+	once.Do(func() {
+		done <- err
+	})
+}
+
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #768, opened 2026-06-04
fix: Concurrent Telegram session creation persists orphan sessions and can clobber current_session

sam-saffron-jarvis:
## What changed

- Stopped `telegramSessionMgr.newSession` from eagerly persisting a Telegram session before it has actually won publication in `m.sessions`.
- Added a small `persistSession` helper and now call it only after the session has been installed as the active runtime for that chat.
- Kept that persistence step under `m.mu` so a same-chat reset cannot race in between publication and `SetCurrent`, which would otherwise let a discarded runtime clobber `metadata.current_session` again.
- Added regression coverage for:
  - concurrent `getOrCreate` on the same chat only persisting the winning session, and
  - stale `resetSessionIfCurrent` calls not leaving behind an extra persisted row or changing the current-session marker.

## Why this is high-value

Telegram updates are handled concurrently, so two quick messages from the same chat can overlap session creation. Before this change, each contender eagerly wrote a fresh `telegram:<chatID>` row and set `current_session` before the winner was known. The losing runtime was closed in memory, but its persisted row remained, and `current_session` could point at that discarded session.

That is a real reliability bug because it fragments persisted chat history, breaks current-session semantics, and can interfere with carryover restore logic that only scans a small recent window of sessions for the chat.

## Validation

- `go build ./...`
- `go test ./...`
- Added targeted regression tests in `internal/serve/telegram_test.go` covering the concurrent creation and stale reset failure modes.

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -518,6 +518,7 @@ func (m *telegramSessionMgr) getOrCreate(ctx context.Context, chatID int64) (*te
 		return existing, nil
 	}
 	m.sessions[chatID] = created
+	m.persistSession(ctx, created)
 	m.mu.Unlock()
 	return created, nil
 }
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #769, opened 2026-06-05
feat: multi-agent gateway with separate-origin widget host (experimental)

arpitjalan:
## Summary

Adds `term-llm serve-gateway` — a thin reverse proxy that fronts the per-agent web serves of several contain workspaces under one origin, and (optionally) serves each agent's widgets from a dedicated, isolated origin.

Screenshot 2026-06-05 at 14 43 47

A `term-llm serve` binds a single agent at startup, so multi-agent access is achieved by fronting many serves rather than routing agents inside one process (already the production shape: one container per agent). The gateway discovers each agent's published port + bearer token from its workspace `.env` and injects the token **server-side**, so the per-agent token never reaches the browser.

This is the foundation for two things: reaching many agents from one place, and the separate-origin widget host that future work (a signed access grant) will use to safely embed/publish widgets.

## What's included

**Agent proxy**
- `GET /agents` (JSON, never includes tokens), an HTML landing page, and `ANY /agent/<name>/...` reverse-proxied to that agent's serve.
- Injects `Authorization: Bearer <token>`; strips client `Authorization`, `Cookie`, `X-Api-Key`, and spoofable `X-Forwarded-*`.
- Rebases the agent's baked-in `/chat` prefix onto `/agent/<name>` so the SPA's API calls, service worker, and subresources all route back through the proxy — zero changes to the agent.

**Separate-origin widget host** (`--widget-host`, default `widgets.localhost`)
- Serves **only** `/w/<agent>/<mount>/...` on a dedicated origin; every other path 404s there. That emptiness is the isolation property: a widget iframe granted `allow-same-origin` is confined to a throwaway origin that hosts nothing sensitive.
- Routes by `Host` header; the agent origin never exposes `/w/`, and the widget origin never exposes `/agents`, `/agent/*`, or the landing page.

**Discovery + UX**
- New `contain port` / `contain token` subcommands and `contain.ReadWebConfig`, the single source for reading a workspace's web config.
- Landing page lists each reachable agent and its widgets (grouped per agent) with links to the widget origin; markup/styles live in embedded `.html`/`.css` files (matching the `serveui` convention).

**Hardening**
- Refuses non-loopback binds (no gateway auth yet).
- Validates `WEB_PORT`; rejects a loopback-literal `--widget-host` that would shadow the gateway; warns at startup if the widget host doesn't resolve.
- Bounded dial/response-header timeouts that leave SSE streams open; per-agent web-config cache invalidated by `.env` mtime; no environment proxy on the backend transport (avoids leaking the injected token via `HTTP_PROXY`).

**Docs:** new `Multi-agent gateway` guide.

## Security posture

Experimental and **loopback-only** by design:
- **No gateway-level authentication yet** — anyone who can reach the gateway can reach every discoverable agent; the enforced loopback bind is the safety net. To expose it, front it with an authenticating proxy and keep the gateway on loopback.
- **No widget access grant yet** — the widget origin proxies any `/w/<agent>/<mount>/` request (loopback only). A signed, expiring HMAC grant for safe cross-origin use is the next step, and the piece that actually closes the widget-iframe XSS in production.

An independent multi-agent code review was run over the diff; the top findings were fixed (HEAD/empty-body response handling, the `HTTP_PROXY` token-leak path, query-string preservation on the widget redirect, and the widget-host/agent-host shadowing guard).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ term-llm
 .cache/*
 .gocache/*
 *.lock
+.DS_Store
--- a/cmd/contain.go
+++ b/cmd/contain.go
@@ -4,7 +4,6 @@ import (
 	"bufio"
 	"fmt"
 	"os"
-	"path/filepath"
 	"sort"
 	"strings"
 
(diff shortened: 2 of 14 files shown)


SamSaffron:
give me a bit arpit, I want to think this one through

arpitjalan:
No rush, thanks Sam!

arpitjalan:
Also, here's some context on scope and the use case driving this:

Why: it comes out of the second-brain Discourse plugin, where widgets are currently proxied through Discourse's own origin and iframed with allow-same-origin - so a prompt-injected widget can run as Discourse (read PMs, act as the user). The fix is to serve widgets from a dedicated, isolated origin, which is what this PR adds (and it doubles as the mechanism for publishing widgets later).

This PR is the foundation: a thin multi-agent reverse proxy + the separate-origin widget host, loopback-only (it refuses non-loopback binds since the agent proxy has no auth yet). Merging it exposes nothing.

Kept intentionally scoped: the widget origin's access control - a short-lived signed grant the embedding app mints and the host verifies, so the Discourse session never reaches the widget origin - is the natural next step, left out here to keep this PR focused on the proxy/isolation foundation.

SamSaffron:
Have not stopped thinking about this... the shape I am coming around to is 

term-llm serve hub

then hub has a registration token and nodes can use that to register on hub, or hub can add nodes it wishes. nodes can live anywhere on docker containers on vms or cloud.

hub UI is to provide a directory and then you can click on any agent to get to the actual UI which is proxied... lots of little details here including ... how do you get back to hub (sidebar), but I am liking it. 

For dv environments this will be great... single pane of glass to see every single dv container, term-llm being so lightweight is a huge advantage here.

continuing my brainstorming but I am making lots of progress here thinking.

arpitjalan:
Closing in favor of https://github.com/SamSaffron/term-llm/commit/f3baaaa477803268cd0ecd51836e1de646ff0775 🎉

PR #770, opened 2026-06-05
fix: WebRTC data-channel requests spawn unbounded handler goroutines

sam-saffron-jarvis:
## What changed

- Bounded WebRTC data-channel request handling to a fixed per-channel concurrency limit instead of spawning an unbounded goroutine for every incoming frame.
- Added a request-slot gate in `runDataChannel()` so once the per-channel limit is reached, the read loop blocks and backpressures the client instead of creating more handler goroutines.
- Added a shutdown-aware stop path so a saturated connection still exits cleanly when the peer is canceled or the idle timeout closes the channel.
- Added unit tests covering both behaviors: blocking when all slots are busy, and prompt unblock when the connection is stopping.

## Why this is high-value

Each WebRTC frame can become a full HTTP request plus `ServeHTTP()` execution. Before this change, a bursty or buggy client could cause `runDataChannel()` to create an unbounded number of goroutines, while writes were still serialized behind a single `sendMu`. That made the transport vulnerable to per-client goroutine and memory blowups, with queued response state piling up behind one writer.

This fix makes the work per connection bounded. At most `maxDataChannelConcurrentRequests` handlers can run at once for a given data channel, and additional frames are naturally backpressured by stopping reads until capacity is available.

## Validation

- Read and verified the current code path in `internal/webrtc/peer.go` did `wg.Add(1)` + `go p.dispatchRequest(...)` for every frame with no bound.
- Ran `gofmt -w internal/webrtc/peer.go internal/webrtc/peer_test.go`
- Ran `go build ./...`
- Ran `go test ./...`
- Ran `git diff --stat`
- Ran `git diff --check`

### DIFF
--- a/internal/webrtc/peer.go
+++ b/internal/webrtc/peer.go
@@ -34,6 +34,10 @@ import (
 
 const maxFrameBytes = 10 * 1024 * 1024 // 10 MB
 
+// Cap per-channel HTTP dispatches so one client cannot create an unbounded
+// number of expensive handler goroutines.
+const maxDataChannelConcurrentRequests = 8
+
 // signalingMsg is a message exchanged via the signaling server.
 type signalingMsg struct {
 	SessionID string `json:"session_id"`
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #771, opened 2026-06-05
fix: Compaction-tail detection does quadratic work every time a compacted session is opened

sam-saffron-jarvis:
## What changed

- Precompute display fingerprints once per `markCompactionDisplayTails()` call instead of rebuilding them inside every duplicate-tail check.
- Replace the length=1..N overlap search in `compactionDuplicateTailRange()` with a single prefix-function scan that finds the longest suffix/prefix match in linear time.
- Add a benchmark covering a large compacted transcript with a duplicated retained tail.

## Why this is high-value

Opening a compacted session runs `LoadScrollbackWithBoundary()`, which calls `markCompactionDisplayTails()` before the TUI can render scrollback. The old implementation did repeated range comparisons for every possible overlap length, which turns duplicate-tail detection into quadratic work on the longest sessions.

This change keeps the existing behavior but reduces the hot-path work to:
- one fingerprint pass over the message list, and
- a linear scan per duplicate-tail candidate.

That cuts both CPU time and repeated comparisons exactly where users feel it most: resuming long-running chats that have already needed compaction.

## Validation

- `gofmt -w internal/session/context_state.go internal/session/context_state_bench_test.go`
- `go build ./...`
- `go test ./...`
- `go test ./internal/session -run '^Test(LoadScrollbackWithBoundary|LoadScrollbackWithBoundaryMarksRetainedRawSuffixHiddenFromDisplay|ApplyCompactionMarksRetainedRawSuffixHiddenFromDisplay)$' -bench BenchmarkMarkCompactionDisplayTailsLargeCompactedTranscript -benchmem`
  - `BenchmarkMarkCompactionDisplayTailsLargeCompactedTranscript-32    614    1860899 ns/op    3539173 B/op    12020 allocs/op`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/session/context_state.go
+++ b/internal/session/context_state.go
@@ -155,6 +155,10 @@ func markCompactionDisplayTails(messages []Message) []Message {
 		return messages
 	}
 	out := append([]Message(nil), messages...)
+	fingerprints := make([]string, len(out))
+	for i := range out {
+		fingerprints[i] = messageDisplayFingerprint(out[i])
+	}
 	for i := 0; i < len(out); i++ {
 		if !isInternalCompactionSummaryMessage(out[i]) {
 			continue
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #772, opened 2026-06-05
fix: Prompt-template git probes can hang session startup because they use uncancelled subprocesses

sam-saffron-jarvis:
## What changed

- Added a shared `runGitOutput` helper in `internal/agents/template.go` that runs git probes with `exec.CommandContext` and a short `gitProbeTimeout`.
- Routed `getGitInfo`, `getGitDiffStat`, and `findGitRoot` through that timeout-bound helper.
- Added focused tests proving template expansion now fails open and returns empty git metadata quickly when git hangs.

## Why this is high-value

Prompt construction runs before any model call starts, and built-in agents commonly use `{{git_branch}}`, `{{git_repo}}`, and `{{git_diff_stat}}`. Before this change, a slow or unhealthy repository/filesystem could block session startup indefinitely because those git subprocesses had no cancellation or timeout. This fix prevents the entire CLI/web/jobs surface from appearing dead by bounding that preflight work and degrading gracefully to empty metadata.

## Validation

- `gofmt -w internal/agents/template.go internal/agents/template_test.go`
- `go build ./...`
- `go test ./...`
- Added timeout regression tests for `{{git_branch}}`/`{{git_repo}}` and `{{git_diff_stat}}`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/agents/template.go
+++ b/internal/agents/template.go
@@ -1,6 +1,7 @@
 package agents
 
 import (
+	"context"
 	"os"
 	"os/exec"
 	"os/user"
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #776, opened 2026-06-07
fix: Memory mining/freshen walks completed sessions with OFFSET pagination, making full scans grow quadratically

sam-saffron-jarvis:
## What changed

- replaced the memory completed-session full walk in `cmd/memory_shared.go` from `OFFSET` pagination to keyset pagination using the last seen session number
- extended `session.ListOptions` with an opt-in number cursor/order path (`BeforeNumber` + `SortByNumberDesc`)
- taught `internal/session/sqlite.go` to use `ORDER BY s.number DESC` plus `INDEXED BY idx_sessions_number` for that path so SQLite can keep scanning from the cursor instead of re-sorting later pages
- added tests covering both paged behavior and the query plan using the session-number index without a temp sort

## Why this is high-value

`memory_mine`, `memory_freshen`, and `memory_status` all call `listCompleteSessions`, which previously paged through the entire completed-session corpus with `LIMIT 200 OFFSET N`.

On SQLite, later `OFFSET` pages must still walk and skip earlier rows, so a full corpus scan gets more expensive as the session table grows. This made long-running memory jobs repay that skip cost over and over.

The new keyset walk turns that into a forward range scan on the existing session-number index:

- page 1: newest completed sessions
- page 2+: `number < last_seen_number`

That removes the repeated skip work and temp sorting from later pages, cutting DB work from repeated rescans to one linear pass over the completed-session set.

## Validation

- verified the issue in current code: `cmd/memory_shared.go` used `Offset`, and `internal/session/sqlite.go` emitted `LIMIT ? OFFSET ?`
- `gofmt -w cmd/memory_shared.go internal/session/types.go internal/session/sqlite.go internal/session/sqlite_test.go`
- `go build ./...`
- `go test ./...`
- added tests:
  - `TestSQLiteStoreListByNumberCursorReturnsCompleteSessions`
  - `TestSQLiteStoreListByNumberCursorUsesSessionNumberIndex`

### DIFF
--- a/cmd/memory_shared.go
+++ b/cmd/memory_shared.go
@@ -21,14 +21,15 @@ func openReadOnlySessionStore(cfg *config.Config) (session.Store, error) {
 
 func listCompleteSessions(ctx context.Context, store session.Store) ([]session.SessionSummary, error) {
 	const pageSize = 200
-	offset := 0
+	var beforeNumber int64
 	all := make([]session.SessionSummary, 0, pageSize)
 
 	for {
 		page, err := store.List(ctx, session.ListOptions{
-			Status: session.StatusComplete,
-			Limit:  pageSize,
-			Offset: offset,
+			Status:           session.StatusComplete,
+			Limit:            pageSize,
+			BeforeNumber:     beforeNumber,
+			SortByNumberDesc: true,
 		})
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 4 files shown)


SamSaffron:
Integrated the completed-session keyset pagination fix into the main working copy; closing this PR.

PR #778, opened 2026-06-08
fix: Fresh session DB initialization always replays every migration

sam-saffron-jarvis:
## What changed

- Detect fresh session databases **before** executing the current schema, so an empty DB is no longer mistaken for a legacy pre-migration DB.
- For the fresh-DB path, install the remaining current-schema objects that were previously only being created as a side effect of replaying migrations (for example `last_user_message_at`, `push_subscriptions`, and current indexes), then record `schema_version` at the latest version immediately.
- Added `TestInitSchemaFreshDBSkipsMigrations`, which uses a sentinel migration to prove that a brand-new DB now takes the fast path instead of replaying the migration stack.

## Why this is high-value

Fresh DB startup was doing the worst possible thing for first-run latency: it executed the full current schema and then still replayed every migration from version 0. That repeated nontrivial work such as table rebuild migrations, backfill `UPDATE`s, and repeated trigger/index setup even though the database had no user data yet.

This change removes that wasted initialization work for:

- first-run local startup
- ephemeral or in-memory stores
- tests creating temporary DBs
- throwaway containers/environments

The result is less SQL work, fewer writes, less schema churn, and lower startup latency on brand-new databases while preserving the existing migration path for real older databases.

## Validation

- `go build ./...`
- `go test ./...`
- Added regression test: `TestInitSchemaFreshDBSkipsMigrations`
- `git diff --check`

### DIFF
--- a/internal/session/sqlite.go
+++ b/internal/session/sqlite.go
@@ -907,6 +907,26 @@ func initSchema(db *sql.DB) error {
 // initSchemaFull handles schema creation and migrations.
 // Only called when schema needs initialization or migration.
 func initSchemaFull(db *sql.DB, versionErr error, currentVersion int) error {
+	noVersionRecord := versionErr != nil && (versionErr == sql.ErrNoRows || strings.Contains(versionErr.Error(), "no such table"))
+	isFreshDB := false
+
+	if noVersionRecord {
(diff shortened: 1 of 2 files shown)


SamSaffron:
Covered by the integrated fresh-DB initialization fix in the main working copy; closing this duplicate PR.

PR #779, opened 2026-06-08
fix: Progressive timeout/cancel can keep the agent running for 5 extra minutes

sam-saffron-jarvis:
## What changed

- changed `progressiveFinalizationContext` so only natural completion gets the detached grace window and 5-minute floor
- timeout/cancel exits now stay attached to the parent context and are capped to the actual reserved finalization budget
- skip finalization entirely when the parent context is already done, preventing post-timeout/cancel synthetic finalization work from starting
- added focused tests for natural vs timeout/cancel finalization behavior and a session-level regression test that proves no detached finalization pass starts after the parent deadline has expired

## Why this is high-value

This fixes a real cancellation/deadline reliability bug in a user-facing execution path. Before this change, a progressive run that had already timed out or been cancelled could still detach itself and keep running a finalization pass for up to 5 more minutes. That meant background jobs could continue spending provider/tool budget, holding resources, and mutating persisted session state after callers had already been told the run stopped.

The fix is intentionally small and preserves the existing “extra polish” behavior for natural completion, while restoring the expected stop guarantees for timeout/cancel paths.

## Validation

- `gofmt -w cmd/progressive.go cmd/progressive_test.go`
- `go build ./...`
- `go test ./...`
- added unit tests covering:
  - natural completion still detaches from parent cancellation
  - timeout finalization uses a reserve-sized budget instead of a 5-minute detached grace window
  - cancelled exits skip finalization when the parent is already cancelled
  - a timed-out progressive session does not start a detached second provider request after the parent deadline expires

### DIFF
--- a/cmd/progressive.go
+++ b/cmd/progressive.go
@@ -530,23 +530,31 @@ func progressiveFinalizeReserve(total time.Duration) time.Duration {
 	return reserve
 }
 
-// progressiveFinalizationContext returns a detached, time-boxed context for the
-// finalization pass plus its CancelFunc, which the caller must invoke (via defer)
-// to release resources. It returns (nil, nil) when there is no grace budget.
+// progressiveFinalizationContext returns a time-boxed context for the finalization
+// pass plus its CancelFunc, which the caller must invoke (via defer) to release
+// resources. Natural completion gets a detached grace window so the model can add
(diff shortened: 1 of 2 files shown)


SamSaffron:
Integrated the progressive timeout/cancel finalization fix into the main working copy; closing this PR.

PR #780, opened 2026-06-08
fix: Fresh session DB initialization always replays the full migration stack

sam-saffron-jarvis:
## What changed

- Detect whether the DB is fresh **before** running `db.Exec(schema)` in `initSchemaFull`, so a brand-new DB no longer gets misclassified as a pre-migration database.
- Bootstrap fresh databases directly at `schemaVersion` instead of replaying every historical migration.
- Add `last_user_message_at` to the base `schema` so fresh installs still get the full current sessions table without depending on replayed migrations.
- Ensure the indexes/constraints that fresh DBs previously got via historical migrations are still created after initialization:
  - `idx_messages_session_sequence`
  - `idx_sessions_status`
  - `idx_sessions_title_skipped`
  - `idx_sessions_last_user_msg`
  - `idx_sessions_last_message`
- Add focused tests that prove:
  - fresh DBs do **not** run historical migrations
  - existing DBs that already have tables but lack `schema_version` **do** still run migrations

## Why this is high-value

Fresh session stores were doing unnecessary startup work on every first run because `initSchemaFull` created the current schema first, then checked for the `sessions` table to decide whether the DB was fresh. That check could never detect a truly empty DB after schema creation, so every fresh DB started at version 0 and replayed the entire migration stack.

This wastes work on exactly the hot path for:

- first-run responsiveness
- `:memory:` / ephemeral stores
- tests
- throwaway containers

As more migrations accumulate, the wasted startup time keeps growing. This fix removes that repeated migration/backfill/rebuild work for brand-new databases while preserving migration behavior for older databases.

## Validation

- Verified the bug first with a focused test: on the pre-fix code, a fresh DB incorrectly ended up running a sentinel migration and recording version `1` instead of `26`.
- Added targeted regression tests in `internal/session/sqlite_test.go` for fresh-vs-existing bootstrap behavior.
- `go build ./...` ✅
- `go test ./internal/session` ✅
- `go test ./...` ⚠️ still fails in unrelated pre-existing `cmd` push-subscription tests:
  - `TestHandlePushSubscribe/POST_saves_subscription`
  - `TestHandlePushSubscribe/DELETE_removes_subscription`
  - `TestPushSubscribe_EndToEnd`
- `git diff --check` ✅

### DIFF
--- a/internal/session/sqlite.go
+++ b/internal/session/sqlite.go
@@ -59,6 +59,7 @@ CREATE TABLE IF NOT EXISTS sessions (
     cwd TEXT,
     created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
     updated_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,
+    last_user_message_at TIMESTAMP,
     last_message_at TIMESTAMP,
     archived BOOLEAN DEFAULT FALSE,
     pinned BOOLEAN DEFAULT FALSE,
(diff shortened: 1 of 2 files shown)


SamSaffron:
Integrated the high-value changes into the main working copy; closing this PR.

PR #789, opened 2026-06-09
fix: Responses API cancel requests can still end as completed

sam-saffron-jarvis:
## What changed

- Updated `responseRun.complete` to give pending cancellation precedence over completion.
- When a cancel request has already been recorded, the run now transitions to `cancelled`, emits `response.cancelled`, and strips completion-only usage fields from the terminal response payload.
- Added a regression test covering the race where `cancelRun()` wins just before `complete()` finalizes the run.

## Why this is high-value

This fixes a user-visible correctness bug in the Responses API: a client could successfully POST `/cancel` and still receive `response.completed` if generation had already finished but the final completion transition had not run yet. That made cancellation flaky on long-running web/API sessions, left the run snapshot looking successful, and could mislead reconnecting SSE clients.

## Validation

- `go build ./...`
- `go test ./...`
- Added `TestResponseRunCompleteHonorsPendingCancellation` to verify cancellation wins over completion.
- `git diff --check`

### DIFF
--- a/cmd/serve_response_runs.go
+++ b/cmd/serve_response_runs.go
@@ -127,6 +127,19 @@ func (r *responseRun) appendEvent(event string, payload map[string]any) error {
 func (r *responseRun) complete(payload map[string]any, usage llm.Usage, sessionUsage llm.Usage) error {
 	r.mu.Lock()
 	defer r.mu.Unlock()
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #792, opened 2026-06-10
fix: serve shutdown cancels detached response runs but never waits for their goroutines to finish

sam-saffron-jarvis:
## What changed

- Added detached-run goroutine tracking to `responseRunManager` with a `sync.WaitGroup`.
- Changed `responseRunManager.Close()` to cancel all tracked runs and then wait for their goroutines to finish before returning.
- Wrapped the detached goroutine launch in `startResponseRun()` with manager-owned tracking, and reject new run starts once shutdown has begun.
- Added a regression test that starts a non-stateful detached response run, blocks its cancellation path, and verifies `Close()` does not return until the run goroutine finishes and stateless runtime cleanup has completed.

## Why this is high-value

Serve shutdown already cancels detached `/v1/responses` runs, but before this change it did not wait for the goroutines spawned by `startResponseRun()` to unwind. Those goroutines own important shutdown-time work for non-stateful runs, including `runtime.Close()` and any final persistence/cleanup done after cancellation. That meant `serve` could continue into store/session teardown while detached runs were still persisting final state or cleaning up providers, creating exactly the kind of restart-time flakiness and state loss that is hard to reproduce.

This fix makes shutdown deterministic: once `serveServer.Stop()` returns from `responseRuns.Close()`, all detached response-run goroutines have finished their teardown.

## Validation

- `gofmt -w cmd/serve_response_runs.go cmd/serve_test.go`
- `go build ./...`
- `go test ./...`
- Added regression test: `TestResponseRunManagerCloseWaitsForDetachedRunShutdown`
- `git diff --check`

### DIFF
--- a/cmd/serve_response_runs.go
+++ b/cmd/serve_response_runs.go
@@ -734,6 +734,7 @@ type responseRunManager struct {
 	activeBySession   map[string]string
 	cleanupTimers     map[string]*time.Timer
 	terminalRetention time.Duration
+	runWG             sync.WaitGroup
 	closed            bool
 }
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing; fix will be applied directly to main.

PR #794, opened 2026-06-10
fix: progressive ask can deadlock when the stream consumer exits early

sam-saffron-jarvis:
## What changed

- made `askProgressiveBridge` shutdown-aware with a dedicated `done` signal
- routed progressive bridge event delivery through a helper that stops returning blocking channel sends once shutdown is signaled
- updated `CloseSuccess` / `CloseError` to avoid blocking when the consumer has already exited
- when the progressive display path returns an error, `ask.go` now signals bridge shutdown before waiting for the producer goroutine, and ignores the expected bridge-closed sentinel from that shutdown path
- added a regression test covering the previously blocking `HandleEvent` and `CloseSuccess` cases after shutdown

## Why this is high-value

`term-llm ask --progressive` runs the LLM/tool producer in one goroutine and drains bridge events in another. Before this change, if the display side exited early (for example a JSON write failure, broken output stream, or renderer failure), the producer kept doing blocking channel sends into the bridge. Once the buffer filled, the producer goroutine wedged forever and the caller hung on `<-runCh`.

This fix makes that failure path unwind cleanly instead of deadlocking a primary user-facing streaming command.

## Validation

- verified the issue in current code by inspecting `cmd/ask_progressive.go`: every bridge event send and close path used plain blocking channel writes with no shutdown signal
- `gofmt -w cmd/ask_progressive.go cmd/ask.go cmd/ask_progressive_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/cmd/ask.go
+++ b/cmd/ask.go
@@ -777,7 +777,13 @@ func runAsk(cmd *cobra.Command, args []string) error {
 			default:
 				err = streamPlainText(displayCtx, events, false)
 			}
+			if bridge != nil && err != nil {
+				bridge.Shutdown()
+			}
 			progressiveRun = <-runCh
+			if err != nil && errors.Is(progressiveRun.Err, errAskProgressiveBridgeClosed) {
+				progressiveRun.Err = nil
+			}
 		}
 		tools.ClearAskUserHooks() // Safe to call even if hooks weren't set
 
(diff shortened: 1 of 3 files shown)


SamSaffron:
Duplicate of #793; fix will be applied directly to main.

PR #797, opened 2026-06-11
fix: Telegram stream persistence can stall event ingestion and trip the watchdog

sam-saffron-jarvis:
## What changed

- moved Telegram response/turn callback persistence off the callback hot path into a small per-stream async queue
- kept callback writes serialized so assistant/tool-result message ordering is preserved
- retained bounded `runStoreOpWithoutCancel` persistence for queued callback writes
- added a regression test that blocks assistant `AddMessage` persistence and verifies the stream still completes instead of timing out before tool progress can continue
- on interrupt/error paths, added a short best-effort drain before fallback persistence so already-queued callback writes get a chance to land first

## Why this is high-value

Telegram streaming callbacks were doing session store writes inline. If assistant persistence stalled inside `AddMessage`/`UpdateMetrics`, the engine could stop advancing before the next tool-progress event reached the Telegram watchdog path. That turns a local persistence stall into a user-visible `stream timed out` failure and truncated replies.

This fix isolates persistence from event ingestion with the smallest targeted change in `internal/serve/telegram.go`, so slow or stuck store writes no longer directly block stream/tool progress.

## Validation

- `go build ./...`
- `go test ./...`
- added `TestStreamReply_CallbackPersistenceDoesNotBlockToolProgress` covering the blocked-persistence failure mode

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -31,6 +31,8 @@ const minEditInterval = 3 * time.Second
 
 // defaultStreamEventTimeout is used when telegramSessionMgr.streamEventTimeout is zero.
 const defaultStreamEventTimeout = 10 * time.Minute
+const telegramCallbackStoreQueueSize = 16
+const telegramCallbackStoreDrainTimeout = 250 * time.Millisecond
 
 const telegramMaxConcurrentHandlers = 8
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #798, opened 2026-06-11
fix: AddMessage still loses messages on concurrent sequence conflicts

sam-saffron-jarvis:
## What changed

- changed `SQLiteStore.AddMessage` so auto-sequenced appends run inside `BEGIN IMMEDIATE` on a pinned connection before reading `MAX(sequence)`
- factored the shared insert/session-timestamp logic into a small helper used by both the auto-sequence and explicit-sequence paths
- added `TestSQLiteStoreAddMessageConcurrentAutoSequence`, a same-session concurrent append stress test that verifies all messages persist with contiguous sequences and no leaked `SQLITE_BUSY` errors

## Why this is high-value

`AddMessage` is on the live persistence path for chat history. Under concurrent appends to the same session, the old deferred transaction could let multiple writers read the same max sequence and then fail during the read→write upgrade window, which I reproduced as escaped `database is locked` / `SQLITE_BUSY` errors. When that happens, a user/assistant/tool message is dropped entirely.

Starting the transaction with `BEGIN IMMEDIATE` moves writer serialization to the start of allocation, so only one writer can reserve the next sequence at a time and the append either waits/retries cleanly or succeeds without losing a message.

## Validation

- `gofmt -w internal/session/sqlite.go internal/session/update_message_test.go`
- `go build ./...`
- `go test ./internal/session -run TestSQLiteStoreAddMessageConcurrentAutoSequence -count=3`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/session/sqlite.go
+++ b/internal/session/sqlite.go
@@ -1787,6 +1787,62 @@ func (s *SQLiteStore) Search(ctx context.Context, opts SearchOptions) ([]SearchR
 	return results, rows.Err()
 }
 
+// addMessageQueryer is the subset of *sql.Tx/*sql.Conn used by AddMessage.
+type addMessageQueryer interface {
+	ExecContext(context.Context, string, ...any) (sql.Result, error)
+	QueryRowContext(context.Context, string, ...any) *sql.Row
+}
+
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #804, opened 2026-06-12
fix: serve shutdown timeout is bypassed by unbounded waits in jobs/response-run/widget teardown

sam-saffron-jarvis:
## What changed

- Made serve shutdown teardown context-aware for jobs, response runs, and widgets by adding bounded close helpers that stop waiting when the serve shutdown context expires.
- Updated `serveServer.Stop` to start those teardown steps in parallel instead of serially, so a slow widget/job/run teardown does not consume the full shutdown window before the HTTP server shutdown begins.
- Added a regression test covering the core failure mode: `serveServer.Stop` now returns on the shutdown deadline even when response-run teardown is blocked.

## Why this is high-value

This fixes a shutdown reliability bug in the long-running `serve` process. Before this change, shutdown advertised a 10 second timeout, but `Stop` could block indefinitely in pre-shutdown teardown because jobs and response runs waited on unbounded `WaitGroup`s and widgets were stopped serially. That could wedge restarts, deploys, watchdog actions, and Ctrl-C handling. With this change, shutdown honors the caller deadline and reaches `server.Shutdown` promptly.

## Validation

- `gofmt -w cmd/serve.go cmd/serve_jobs_v2.go cmd/serve_response_runs.go cmd/serve_stop_test.go internal/widgets/manager.go`
- `go build ./...`
- `go test ./...`
- Added `TestServeStopHonorsContextWhenResponseRunCloseBlocks`
- `git diff --check`

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -1118,6 +1118,38 @@ func (s *serveServer) contextWithShutdown(ctx context.Context) (context.Context,
 	return ctx, cancel
 }
 
+func runServeStopTasks(ctx context.Context, tasks ...func(context.Context) error) <-chan error {
+	done := make(chan error, 1)
+	go func() {
+		if len(tasks) == 0 {
+			done <- nil
+			return
+		}
+		var wg sync.WaitGroup
+		errCh := make(chan error, len(tasks))
+		for _, task := range tasks {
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #805, opened 2026-06-12
fix: serve shutdown timeout is bypassed by unbounded waits in manager teardown

sam-saffron-jarvis:
## What changed

- started `serveServer.Stop` shutdown and manager teardown concurrently so `server.Shutdown(ctx)` is no longer blocked behind synchronous subsystem cleanup
- added context-aware teardown paths for `jobsV2Manager`, `responseRunManager`, and `widgets.Manager`
- kept the existing blocking `Close()` methods by having them call the new context-aware variants with `context.Background()`
- added a regression test that proves `serveServer.Stop` now returns on the shutdown deadline even when manager waitgroups never finish

## Why this is high-value

This fixes a real shutdown reliability bug: the outer serve loop advertises a 10-second shutdown budget, but before this change `Stop` could block indefinitely in `jobsV2.Close()`, `responseRunManager.Close()`, or serial widget teardown and never even reach `server.Shutdown(ctx)`. That can wedge deploys, restarts, watchdog recovery, and Ctrl-C on long-running services.

After this change, shutdown still requests normal cleanup first, but the waits are bounded by the caller's context so the process can reliably stop within the configured budget.

## Validation

- `gofmt -w cmd/serve.go cmd/serve_jobs_v2.go cmd/serve_response_runs.go cmd/serve_shutdown.go cmd/serve_test.go internal/widgets/manager.go`
- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -1129,24 +1129,67 @@ func (s *serveServer) Stop(ctx context.Context) error {
 			close(s.shutdownCh)
 		}
 	})
-	if s.jobsV2 != nil {
-		_ = s.jobsV2.Close()
-	}
-	if s.responseRuns != nil {
-		s.responseRuns.Close()
-	}
-	if s.widgetsMgr != nil {
-		s.widgetsMgr.Close()
-	}
-	closeFileTrackingStore()
-	s.modelsMu.Lock()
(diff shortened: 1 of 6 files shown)


SamSaffron:

PR #806, opened 2026-06-12
fix: Sidebar session listing does a full sort on every web refresh because the ORDER BY cannot use any existing index

sam-saffron-jarvis:
## What changed

- Added two dedicated SQLite expression indexes for the sidebar list orders:
  - `idx_sessions_sidebar_activity` on `(archived, COALESCE(pinned, FALSE) DESC, COALESCE(last_message_at, last_user_message_at, created_at) DESC)`
  - `idx_sessions_sidebar_last_user_activity` on `(archived, COALESCE(pinned, FALSE) DESC, COALESCE(last_user_message_at, created_at) DESC)`
- Added schema migration 30 plus startup index creation so both upgraded and fresh databases get the new indexes.
- Switched `List()` to use the same unqualified ORDER BY expressions as the new indexes so SQLite can match them exactly.
- Added query-plan tests proving the activity-sorted and last-user-activity-sorted queries use the new indexes without a temp sort.

## Why this is high-value

The web sidebar hits `store.List(... SortByActivity: true, Limit: 100)` on refresh/poll. Before this change, SQLite had no index matching `archived + pinned + COALESCE(...) ORDER BY`, so the planner did a scan plus `USE TEMP B-TREE FOR ORDER BY` for each refresh. That work grows with session count and directly affects a user-visible UI path.

With the new indexes, SQLite can read the top sessions in index order and stop at the limit, avoiding the full candidate sort and reducing CPU/latency on a hot path.

## Validation

- Verified the issue was real first with `EXPLAIN QUERY PLAN`, which showed:
  - `SCAN s`
  - `USE TEMP B-TREE FOR ORDER BY`
- Added tests that assert the two sidebar ORDER BY variants use:
  - `idx_sessions_sidebar_activity`
  - `idx_sessions_sidebar_last_user_activity`
  and do **not** use a temp B-tree sort.
- Ran:
  - `go build ./...`
  - `go test ./...`
  - `git diff --check`

### DIFF
--- a/internal/session/sqlite.go
+++ b/internal/session/sqlite.go
@@ -234,7 +234,15 @@ func NewSQLiteStore(cfg Config) (*SQLiteStore, error) {
 // - Fresh databases get the full schema from `schema` const and start at this version
 // - Existing databases run migrations to reach this version
 // Increment when adding new migrations.
-const schemaVersion = 29
+const schemaVersion = 30
+
+const (
+	// Keep these expressions aligned with List()'s hot ORDER BY clauses so SQLite
+	// can satisfy the sidebar top-N query directly from the matching indexes.
+	sessionListPinnedOrderExpr       = "COALESCE(pinned, FALSE)"
+	sessionListActivityOrderExpr     = "COALESCE(last_message_at, last_user_message_at, created_at)"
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #819, opened 2026-06-13
fix: Telegram session shutdown can block forever waiting for a stuck stream

sam-saffron-jarvis:
## What changed

- changed Telegram session shutdown to use a bounded wait instead of blocking forever on `replyDone`
- reused the manager timeout setting for this shutdown wait, with a 5s fallback when no timeout is configured
- logged when a stuck stream does not stop in time and continued best-effort cleanup
- added a regression test covering `resetSessionIfCurrent` when the active stream ignores cancellation and never closes `replyDone`

## Why this is high-value

A single wedged Telegram stream could previously block `/reset`, session replacement, or full Telegram platform shutdown forever because `closeTelegramSession` waited on `replyDone` with no timeout. That can stall operational recovery and service reload/shutdown. This change converts that failure mode into a bounded wait with logging, so the manager can keep making progress even when a provider or tool path is stuck.

## Validation

- `go test ./internal/serve -run 'TestTelegramSessionMgrResetSessionIfCurrent_(CancelsActiveStream|TimesOutStuckStreamShutdown)$'`
- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -515,7 +515,7 @@ func (m *telegramSessionMgr) getOrCreate(ctx context.Context, chatID int64) (*te
 	m.mu.Lock()
 	if existing, ok := m.sessions[chatID]; ok {
 		m.mu.Unlock()
-		closeTelegramSession(created)
+		m.closeTelegramSession(created)
 		return existing, nil
 	}
 	m.sessions[chatID] = created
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented in the working tree: Telegram session shutdown now uses a bounded wait for stuck reply streams, with regression coverage.

PR #821, opened 2026-06-13
fix: Widget processes can start after shutdown begins and leak past CloseContext

sam-saffron-jarvis:
## What changed

- added a manager shutdown flag and shutdown context so widget startup can see when `CloseContext()` has begun
- reject new widget starts in `ensureRunning()` once shutdown starts, including requests that were already waiting on `startMu`
- run widget commands with the manager shutdown context and re-check shutdown state before/during/after startup so a process cannot survive a late startup race
- added a regression test covering the race where shutdown starts while a request is queued to start a widget

## Why this is high-value

This closes a real shutdown race where `CloseContext()` took a one-time snapshot of current widget entries, but a concurrent request could still start a widget process after that snapshot. Because shutdown only performs one stop pass, that late-started process could outlive the server, keep ports or sockets bound, and break later restarts. Preventing orphaned subprocesses during shutdown/reload is a direct reliability win.

## Validation

- `gofmt -w internal/widgets/manager.go internal/widgets/manager_test.go`
- `go build ./...`
- `go test ./...`
- added `TestManagerCloseContextPreventsStartAfterShutdownBegins`
- `git diff --check`

### DIFF
--- a/internal/widgets/manager.go
+++ b/internal/widgets/manager.go
@@ -2,6 +2,7 @@ package widgets
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"net"
@@ -23,6 +24,8 @@ const (
 	startupTimeout   = 10 * time.Second
 	socketRuntimeDir = "/tmp/term-llm-widgets"
 )
+
+var errManagerClosed = errors.New("widget manager is shutting down")
 
 type processState int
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented in the working tree: widget manager shutdown now rejects/cleans up late starts after CloseContext begins, with regression coverage.

PR #822, opened 2026-06-13
fix(claude-bin): frame replayed history so resumed sessions don't restart

Flink:
## Summary

When a `claude-bin` runtime is recreated — runtime eviction after idle, or a
server restart — the Claude CLI `--resume` state is gone. On the next turn the
provider therefore re-submits the **whole transcript** from scratch. Both stdin
paths mishandled this replay, causing a resumed conversation to be re-answered
**from the very beginning**: the model re-did its opening actions, replied again
to user messages it had already answered, and repeated things like earlier
apologies.

## Root cause

`claude-bin` is the Claude Code CLI, not a stateless API. Continuity is kept in
memory via `sessionID` (the CLI session) + `messagesSent`. A fresh provider
instance (`sessionID == ""`) replays everything, and the two builders both lose
turn structure:

- **Text mode** (`buildConversationPrompt`): flattened the entire transcript into
  a single `User: … / Assistant: …` blob delivered as **one pasted user turn**.
  Claude read it as a freshly-pasted thread starting at the first `User:` line and
  answered it top-to-bottom.
- **stream-json mode** (`buildStreamJsonInput`, selected when the history carries
  an image): only ever emitted **user-role** turns and **dropped every assistant
  reply entirely**. Claude received a pile of unanswered questions and re-answered
  all of them — strictly worse than text mode.

This only manifests on **resume** (a new runtime); within a live runtime,
follow-up turns use `--resume` and keep proper turn structure, so it looked fine
until you reopened an old conversation.

## Fix

Split the message slice at the latest user turn. When genuine prior history
precedes it, wrap the earlier turns (assistant replies **included**) in a
`<conversation_history>` block followed by an explicit instruction
(`claudeBinResumeReplayInstruction`) telling Claude Code those turns are already
handled and to reply only to the latest user message. The latest turn is still
emitted live — in stream-json mode it stays a real stream-json turn, so any image
in the current message still reaches the model as a true vision block.

Shared helpers used by both paths:
- `conversationFinalTurnStart` — index where the latest user turn begins (folds an
  immediately preceding developer message into that turn).
- `messagesContainPriorAssistantTurn` — gates the framing on a prior **assistant**
  turn. A lone tool-result image with no preceding assistant turn is deliberately
  *not* framed, so the synthetic tool-result image replay path keeps emitting a
  live vision turn. (Tools are always invoked by the assistant, so any real
  multi-turn history carries an assistant turn — this never misses a genuine
  transcript.)
- `renderConversationParts` — extracted transcript renderer, now returning
  `[]string` so callers can slice history vs. final turn.

## Behavior matrix

| Situation | Before | After |
|---|---|---|
| First turn / single user message | as-is | unchanged (no framing) |
| Live runtime follow-up (`--resume`) | only new msg sent | unchanged |
| Resume, text history | whole thread re-answered | framed history + reply to last only |
| Resume, history with image (stream-json) | assistant turns dropped, all re-answered | framed history (assistant replies kept) + last turn live |
| Trailing tool-result image, no prior assistant | live vision turn | unchanged (not framed) |

## Tests

Added regression tests for both stdin paths:
- `TestBuildConversationPrompt_WrapsReplayedHistory`
- `TestBuildConversationPrompt_SingleTurnNoFraming`
- `TestBuildStreamJsonInput_FramesPriorHistoryAndKeepsFinalImage`
- `TestBuildStreamJsonInput_NoFramingWithoutPriorTurn`

Existing `claude_bin` tests (incl. the tool-result-image ordering test) still
pass; `gofmt`, `go vet ./internal/llm/` and `go test ./internal/llm/` are green.

## Verification

Reproduced on a real resumed web conversation (history containing an image →
stream-json path) and confirmed live that the agent now continues from the latest
message instead of replaying the thread.

### DIFF
--- a/internal/llm/claude_bin.go
+++ b/internal/llm/claude_bin.go
@@ -1506,7 +1506,86 @@ func (p *ClaudeBinProvider) systemPromptForTurn(messages []Message) string {
 
 // buildConversationPrompt constructs the conversation prompt from messages.
 // This can be called with a subset of messages when resuming a session.
+//
+// When the slice carries a multi-turn history (earlier assistant/tool turns
+// precede the latest user message), the earlier turns are wrapped in a
+// <conversation_history> block with an explicit instruction so Claude Code
+// treats them as already-handled context and replies only to the latest user
+// message. Without this, a fresh CLI session — e.g. after a runtime eviction or
+// server restart, when --resume state is gone — receives the whole transcript
+// as a single pasted user turn and re-answers it from the very beginning.
 func (p *ClaudeBinProvider) buildConversationPrompt(messages []Message) string {
+	finalTurnStart := conversationFinalTurnStart(messages)
+	if finalTurnStart <= 0 || !messagesContainPriorAssistantTurn(messages[:finalTurnStart]) {
+		return strings.TrimSpace(strings.Join(p.renderConversationParts(messages), "\n\n"))
+	}
+
+	historyText := strings.TrimSpace(strings.Join(p.renderConversationParts(messages[:finalTurnStart]), "\n\n"))
+	finalText := strings.TrimSpace(strings.Join(p.renderConversationParts(messages[finalTurnStart:]), "\n\n"))
+	switch {
+	case historyText == "":
+		return finalText
+	case finalText == "":
+		return historyText
+	}
+
+	var b strings.Builder
+	b.WriteString("<conversation_history>\n")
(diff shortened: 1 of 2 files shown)


SamSaffron:
oh I think I saw this, thanks so much for chasing it down.

SamSaffron:
will do a big followup I think the biggest gap is that our session store is not storing the claude session id in sqlite

PR #824, opened 2026-06-13
perf: render streaming markdown blocks incrementally

sam-saffron-jarvis:
## What

Adds a conservative fast path for streaming markdown committed blocks:

- renders newly committed blocks incrementally instead of re-rendering the full accumulated markdown when the block is locally safe
- preserves the existing full-document render as the fallback/oracle for globally-sensitive markdown
- disables the fast path after reference-style/global constructs that can change earlier render output
- keeps resettable/non-resettable snapshot handling unchanged via `applyRenderedSnapshot`

Also adds a streaming markdown corpus and chunking parity harness:

- direct render vs streamed render
- all-at-once, byte-by-byte, line-by-line, fixed chunk sizes, random deterministic chunks
- explicit 100-piece splits
- adversarial cuts around markdown syntax bytes
- corpus covers headings, paragraphs, inline formatting, code fences, setext headings, rules, lists, blockquotes, tables, reference links, tabs, and cursed split boundaries

No session transcripts or private examples were committed; the corpus is synthetic.

## Why

The previous streaming renderer preserved correctness by re-rendering all accumulated markdown on each committed block. That is safe but has an accidental quadratic shape for long streamed responses.

The existing benchmark showed:

```text
BenchmarkStreamRendererLargeNoTabs
~63.3 ms/op
~38.5 MB/op
~357k allocs/op
```

This PR keeps the safety fallback while adding a fast path for the common case.

After:

```text
BenchmarkStreamRendererLargeNoTabs
~2.8-2.9 ms/op
~3.4 MB/op
~17k allocs/op
```

That is roughly:

- ~22x faster wall time
- ~11x less allocated memory
- ~21x fewer allocations

## Safety

The optimization is deliberately conservative:

- full render remains the fallback
- global/reference-style markdown disables the incremental path
- final output is protected by corpus parity tests across hostile chunking patterns
- existing CommonMark and streaming tests still pass

Verification run:

```text
go test ./... -count=1
go build ./...
git diff --check
go test ./internal/ui/streaming -run '^$' -bench 'BenchmarkStreamRendererLargeNoTabs|BenchmarkStreamingVsDirect|BenchmarkStreamingFull|BenchmarkStreamingChunked' -benchmem -count=5
```

### DIFF
--- a/internal/ui/streaming/committed_parity_test.go
+++ b/internal/ui/streaming/committed_parity_test.go
@@ -1,0 +1,111 @@
+package streaming
+
+import (
+	"bytes"
+	"testing"
+)
+
+func TestCommittedRenderedMatchesDirectAtStreamingBoundaries(t *testing.T) {
+	tests := []struct {
+		name    string
+		input   string
+		partial bool
+	}{
+		{
+			name:  "positive incremental paragraphs and heading",
+			input: "Alpha.\n\n## Heading\n\nBeta.\n\n```go\nfmt.Println(\"hi\")\n```\n",
+		},
+		{
+			name:    "positive incremental production partial mode",
+			input:   "Alpha.\n\n## Heading\n\nBeta.\n\n```go\nfmt.Println(\"hi\")\n```\n",
+			partial: true,
+		},
+		{
+			name:  "tight list sibling flush stays tight",
+			input: "- a\n- b\n- c\n",
+		},
+		{
+			name:  "indented heading-looking line remains paragraph continuation",
+			input: "foo\n    # bar\n",
+		},
+		{
+			name:  "indented thematic-looking line remains paragraph continuation",
+			input: "Foo\n    ***\n",
+		},
+		{
+			name:  "blockquote reference definition can rewrite earlier shortcut ref",
+			input: "[foo]\n\n> [foo]: https://example.com\n\nAfter\n",
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			sr := newCommittedParityRenderer(t, &bytes.Buffer{}, tt.partial)
+			for i := 0; i < len(tt.input); i++ {
+				if _, err := sr.Write([]byte{tt.input[i]}); err != nil {
+					t.Fatalf("write byte %d failed: %v", i, err)
+				}
+				assertCommittedRenderedMatchesDirect(t, sr, []byte(tt.input[:i+1]))
+			}
+		})
+	}
+}
+
+func TestCorpusCommittedRenderedMatchesDirectAtStreamingBoundaries(t *testing.T) {
+	for _, doc := range loadCorpusDocuments(t) {
+		for _, partial := range []bool{false, true} {
+			name := doc.name
+			if partial {
+				name += "/partial"
+			} else {
+				name += "/plain"
+			}
+			t.Run(name, func(t *testing.T) {
+				sr := newCommittedParityRenderer(t, &bytes.Buffer{}, partial)
+				for i := 0; i < len(doc.content); i++ {
+					if _, err := sr.Write(doc.content[i : i+1]); err != nil {
+						t.Fatalf("write byte %d failed: %v", i, err)
+					}
+					assertCommittedRenderedMatchesDirect(t, sr, doc.content[:i+1])
+				}
+			})
+		}
+	}
+}
+
+func newCommittedParityRenderer(t testing.TB, w *bytes.Buffer, partial bool) *StreamRenderer {
+	t.Helper()
+	if partial {
+		sr, err := NewRendererWithOptions(
+			w,
+			newTestMarkdownRenderer(testRenderWidth),
+			[]StreamRendererOption{WithPartialRendering()},
+		)
+		if err != nil {
+			t.Fatalf("NewRendererWithOptions failed: %v", err)
+		}
+		return sr
+	}
+	sr, err := NewRenderer(w, newTestMarkdownRenderer(testRenderWidth))
+	if err != nil {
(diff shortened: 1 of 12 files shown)


sam-saffron-jarvis:
Added fuzz coverage to the PR after review.

Fuzz target: `FuzzStreamingCorpusParity`

It seeds from the synthetic corpus plus focused markdown snippets, then mutates arbitrary UTF-8 inputs up to 4KB and compares normalized direct render against streamed render under multiple chunking modes:

- all-at-once
- byte-by-byte
- 100-piece split
- deterministic random chunks
- adversarial markdown-boundary cuts

Run results:

```text
go test ./internal/ui/streaming -run FuzzStreamingCorpusParity -count=1
ok

go test ./internal/ui/streaming -run "^$" -fuzz FuzzStreamingCorpusParity -fuzztime=60s
PASS, 776,086 execs

go test ./internal/ui/streaming -run "^$" -fuzz FuzzStreamingCorpusParity -fuzztime=180s
PASS, 6,223,201 execs
```

Also re-ran:

```text
go test ./... -count=1
go build ./...
git diff --check
```

All passed.

sam-saffron-jarvis:
Added CommonMark-specific coverage too.

New pieces:

- `TestCommonMarkFullSpecMarkdownStreamingParity`
  - combines all 652 CommonMark markdown examples into one synthetic markdown document (~16KB)
  - compares normalized direct render vs streamed render under:
    - all-at-once
    - 100-piece split
    - random chunks
    - adversarial markdown-boundary cuts

- `FuzzCommonMarkFullSpecChunking`
  - keeps the full combined CommonMark markdown document fixed
  - fuzzes seed/mode to vary chunking across 100-piece/random/adversarial modes

- `FuzzStreamingCorpusParity` now also seeds every individual CommonMark markdown example, plus the synthetic local corpus.

Runs:

```text
go test ./internal/ui/streaming -run "TestCommonMarkFullSpecMarkdownStreamingParity|FuzzStreamingCorpusParity|FuzzCommonMarkFullSpecChunking" -count=1
ok

go test ./internal/ui/streaming -run "^$" -fuzz FuzzStreamingCorpusParity -fuzztime=60s
PASS, 4,356,417 execs

go test ./internal/ui/streaming -run "^$" -fuzz FuzzCommonMarkFullSpecChunking -fuzztime=60s
PASS, 763 full-spec chunking executions
```

I also found one pathological generated fuzz input with ~3.5KB of blockquote markers that made the underlying markdown renderer take ~47s. It reproduced as slow-but-passing and is not caused by this PR; I capped the arbitrary-input fuzz target at 2KB and kept full CommonMark fuzzing in the dedicated fixed-document chunking target. Then re-ran:

```text
go test ./... -count=1
go build ./...
git diff --check
benchmarks count=3
```

All passed.

sam-saffron-jarvis:
Ran the requested `chatgpt:gpt-5.5-xhigh` review. It found real issues in the previous version: final `Close()` parity masked incorrect intermediate committed snapshots for lists / approximate block detection / reference definitions.

I amended the PR to address that:

- narrowed the fast path substantially
  - lists, tables, blockquotes, and paragraph-interruption commits now use full-document render
  - fenced code only fast-paths when it is top-level/CommonMark-valid indentation
  - headings/thematic breaks only fast-path when indentation cannot mean indented-code continuation
- strengthened global markdown detection
  - reference definitions disable future fast path
  - shortcut/reference-style link candidates are treated as unsafe unless inline-link-shaped
- added `TestCommittedRenderedMatchesDirectAtStreamingBoundaries`
  - checks `CommittedRendered()` against direct render after every streamed byte for review repros:
    - tight list sibling flush stays tight
    - indented heading-looking line remains paragraph continuation
    - indented thematic-looking line remains paragraph continuation
    - blockquote reference definition can rewrite earlier shortcut ref

Verification after fixes:

```text
go test ./... -count=1
go build ./...
git diff --check

go test ./internal/ui/streaming -run "^$" -fuzz FuzzStreamingCorpusParity -fuzztime=60s
PASS, 4,753,836 execs

go test ./internal/ui/streaming -run "^$" -fuzz FuzzCommonMarkFullSpecChunking -fuzztime=30s
PASS, 364 full-spec chunking execs
```

Bench after narrowing is still good for the original hotspot:

```text
BenchmarkStreamRendererLargeNoTabs
~2.88–3.01 ms/op
~3.42 MB/op
~17,075 allocs/op
```

So we keep the main ~20x win while removing the unsafe intermediate-commit behavior the reviewer caught.

sam-saffron-jarvis:
Follow-up from the `claude-bin:opus-max` review is now pushed.

What changed since the previous head:

- Added positive fast-path committed-boundary coverage:
  - simple paragraphs/headings/fenced code
  - same case under `WithPartialRendering()` to mirror production
- Expanded committed-boundary oracle across the whole synthetic corpus in both plain and partial modes:
  - after each streamed byte, compare `CommittedRendered()` to direct render of `input[:CommittedMarkdownLen()]`
- Made `assertRenderedEqual` accept `testing.TB` so fuzz/tests can share the oracle
- Added a comment tying the incremental `\n\n` separator to the ANSI renderer top-level separator, with committed-boundary tests as the guard

Re-ran:

```text
go test ./internal/ui/streaming -count=1
go test ./... -count=1
go build ./...
git diff --check
benchmarks count=3
```

All passed.

The attempted `venice:zai-org-glm-5-1-max` review ran for ~5 minutes then worker-lost. Its partial output repeated the same final-output-vs-committed-output oracle concern from opus, which this push addresses. It also mentioned unrelated schema/scope changes, but current PR diff is only `internal/ui/streaming/*`; those comments do not apply to this branch.

SamSaffron:
Applied locally to main working copy for follow-up.

PR #826, opened 2026-06-14
fix: TUI drops visible partial assistant replies on stream error/cancel

sam-saffron-jarvis:
## What changed

- salvage visible partial assistant output when a TUI stream ends with `StreamEventError` or cancellation
- synthesize a fallback assistant message from the accumulated streamed text/reasoning, append it to in-memory `m.messages`, and invalidate history/context caches
- best-effort persist that fallback with a short timeout, updating the existing pending assistant row when one already exists so interrupted streams do not create duplicate assistant rows
- add regression tests covering both the fresh-fallback path and the pending-row update path

## Why this is high-value

This fixes a real data-loss path in the main interactive TUI surface. Before this change, users could see partial assistant text on screen after an interrupted or failed stream, but that visible content was not added to conversation state and could be missing from persisted session history. That meant the next turn could lose context, and resuming the session later could lose the partial reply entirely.

## Validation

- `gofmt -w internal/tui/chat/chat.go internal/tui/chat/streaming_test.go`
- `go build ./...`
- `go test ./...`
- added tests:
  - `TestUpdate_StreamErrorSalvagesPartialAssistantReply`
  - `TestUpdate_StreamErrorUpdatesPendingAssistantFallbackRow`
- `git diff --check`

### DIFF
--- a/internal/tui/chat/chat.go
+++ b/internal/tui/chat/chat.go
@@ -1047,6 +1047,67 @@ func (m *Model) flushStreamingContentOnErrorToScrollback() []tea.Cmd {
 	return ui.ScrollbackPrintlnCommands(output, true)
 }
 
+func (m *Model) interruptedAssistantFallbackMessage() (llm.Message, bool) {
+	responseContent := m.currentResponse.String()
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #828, opened 2026-06-14
fix: Agent on_complete hooks can hang ask forever and leak subprocesses

sam-saffron-jarvis:
## What changed

- wrapped `runOnCompleteCapture` in a bounded `context.WithTimeout` so `ask` no longer waits forever on a stuck `agent.on_complete` hook
- switched `on_complete` execution to `exec.CommandContext` plus `tools.PrepareCommand`, which applies the same process-group cancellation/cleanup used by other command execution paths
- set `WaitDelay` for `on_complete` so `ask` also returns when the parent shell exits but a backgrounded child keeps stdout/stderr open
- added focused tests covering both a hung hook timeout and the background-child/stdout-pipe hang case

## Why this is high-value

`agent.on_complete` runs after the model has already successfully finished. Before this change, a slow or misbehaving hook could still leave `term-llm ask` hung indefinitely at the very end, which is especially painful for scripts and automation. The old path also had no process-group cleanup, so broken hooks could leave stray subprocesses behind.

This fix makes that tail-end hook fail open: the main `ask` result still completes, while timeout/failure continues to surface as the existing non-fatal warning.

## Validation

- `gofmt -w cmd/ask.go cmd/ask_test.go`
- `go build ./...`
- `go test ./cmd -run 'TestRunOnCompleteCapture_'`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/cmd/ask.go
+++ b/cmd/ask.go
@@ -2601,6 +2601,11 @@ type onCompleteResult struct {
 	Stderr string
 }
 
+var (
+	onCompleteTimeout   = 10 * time.Second
+	onCompleteWaitDelay = 2 * time.Second
+)
+
 // runOnCompleteCapture executes the on_complete shell command with input piped to stdin
 // and captures stdout/stderr. Runs in the git repo root if available, else cwd.
 func runOnCompleteCapture(command, input string) (onCompleteResult, error) {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #830, opened 2026-06-15
fix: TUI cancel/quit clears stream persistence state before the stream actually exits

sam-saffron-jarvis:
## What changed

- Stopped the Ctrl+C / Esc cancellation paths from clearing stream callbacks and pending assistant persistence state before the engine stream goroutine exits.
- Added a `streamCancelRequested` state so user-initiated cancellation stays in a stopping state until shutdown finishes.
- Taught `listenForStreamEventsSync()` to synthesize `context.Canceled` when a canceled stream closes silently, so the existing interrupted-error cleanup path runs instead of the normal success path.
- Covered the regression with tests for both Esc cancellation state preservation and canceled-stream shutdown cleanup.

## Why this is high-value

This fixes a real transcript-corruption race on the main TUI path.

Before this change, cancel/quit cleared `pendingAssistantMsgID`, `pendingAssistantSnapshot`, and related persistence state immediately after calling `streamCancelFunc()`, even though the engine stream goroutine could still run already-captured callbacks while finishing. That could:

- lose the partial assistant snapshot needed by interrupted-message salvage
- fall back from updating the pending assistant row to inserting a new one
- misclassify a canceled stream as a successful completion when the adapter channel closed without an explicit error

The result was user-visible dropped partial replies or duplicate/misordered assistant rows after cancellation.

## Validation

- `gofmt -w internal/tui/chat/chat.go internal/tui/chat/handlers.go internal/tui/chat/streaming.go internal/tui/chat/handlers_test.go internal/tui/chat/streaming_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/tui/chat/chat.go
+++ b/internal/tui/chat/chat.go
@@ -94,15 +94,16 @@ type Model struct {
 	reasoningRawWarned       bool
 
 	// Streaming state
-	currentResponse  strings.Builder
-	currentTokens    int
-	streamStartTime  time.Time
-	webSearchUsed    bool
-	retryStatus      string
-	streamCancelFunc context.CancelFunc
-	streamDone       chan struct{}       // closed when the engine goroutine exits
-	tracker          *ui.ToolTracker     // Tool and segment tracking (shared component)
-	subagentTracker  *ui.SubagentTracker // Subagent progress tracking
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #831, opened 2026-06-16
fix: wait_for_jobs can return the wrong run because queue_agent discards the triggered run ID

sam-saffron-jarvis:
## What changed

- queue_agent now preserves the run ID returned by the jobs-v2 trigger endpoint and includes it in its JSON result alongside the job ID
- wait_for_jobs now accepts `run_ids` and, when provided, polls `/v2/runs/{run_id}` so it waits on the exact triggered run instead of whichever run is latest for the job
- kept `job_ids` support for backwards compatibility and added a focused test covering the exact-run polling path

## Why this is high-value

This fixes a user-visible reliability bug on the queued sub-agent path. Before this change, if the same queued job was triggered again before wait_for_jobs finished polling, the waiter could attach to the newer run and return the wrong status/output. Returning and consuming the trigger's run ID removes that ambiguity and prevents result corruption.

## Validation

- `gofmt -w internal/tools/queue_agent.go internal/tools/queue_agent_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/tools/queue_agent.go
+++ b/internal/tools/queue_agent.go
@@ -33,12 +33,14 @@ type QueueAgentArgs struct {
 
 type QueueAgentResult struct {
 	JobID     string `json:"job_id"`
+	RunID     string `json:"run_id"`
 	AgentName string `json:"agent_name"`
 }
 
 type WaitForJobsArgs struct {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #834, opened 2026-06-17
fix: WebRTC data-channel requests keep running after disconnect and can hang peer cleanup

sam-saffron-jarvis:
## What changed

- created a data-channel-scoped request context in `runDataChannel` and cancel it before waiting for in-flight handlers during disconnect, idle-timeout, and reader-exit teardown paths
- made data-channel send failures stop the channel so all in-flight WebRTC requests get canceled promptly
- updated `dispatchRequest` to create a per-request child context and taught `dcResponseWriter` to fail closed by canceling that request context on the first send error
- added a regression test proving a handler blocked on `r.Context().Done()` is released when response-frame delivery fails

## Why this is high-value

A closed browser tab or broken data channel could previously leave long-lived WebRTC handlers running under the peer's long-lived context, while peer teardown blocked in `wg.Wait()`. That can strand streaming `/v1/responses` work, model runs, and goroutines until the handler exits on its own. This fix makes disconnects and transport write failures actively cancel request contexts, so cleanup and reconnect behavior are reliable instead of waiting on orphaned handlers.

## Validation

- `gofmt -w internal/webrtc/peer.go internal/webrtc/peer_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/webrtc/peer.go
+++ b/internal/webrtc/peer.go
@@ -510,17 +510,14 @@ func (p *peer) handleOffer(ctx context.Context, offer signalingMsg) {
 }
 
 func (p *peer) runDataChannel(ctx context.Context, dc *datachannel.DataChannel) {
+	dataChannelCtx, cancelDataChannelRequests := context.WithCancel(ctx)
+	defer cancelDataChannelRequests()
+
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #835, opened 2026-06-17
fix: serve runtime persistence callbacks bypass engine deadlines and can stall streams/tool execution

sam-saffron-jarvis:
## What changed

- Added an `inlinePersistContext` helper in `cmd/serve_runtime.go` that keeps the incoming context's cancellation and deadline while still bounding inline DB writes.
- Switched the callback-driven persistence paths to use that preserved context for:
  - `persistSnapshot`
  - `appendMessages`
  - the in-progress assistant upsert inside `upsertPendingAssistantLocked`
- Left the existing cancel-proof deferred/final reconciliation path unchanged, so best-effort post-cancel cleanup still uses its own detached timeout after the engine is done.
- Added tests that verify inline persistence keeps short parent deadlines and that the snapshot callback's `AddMessage` path stays within the engine callback-scale deadline.

## Why this is high-value

These store writes run inline on the engine's streaming/tool orchestration path. Before this change, they wrapped the callback context with `context.WithoutCancel(..., 10s)`, which discarded the engine's 5s callback timeout and could let a slow or locked session DB stall streaming and delay tool execution for up to 10 seconds per callback.

Preserving the callback context prevents that deadline bypass. The exact failure avoided is: a slow SQLite write in assistant snapshot/response/turn persistence no longer outlives the engine's callback budget and freeze the active run while waiting on DB I/O.

## Validation

- Verified the issue is real in current code by inspecting `internal/llm/engine.go` and `cmd/serve_runtime.go`: engine callbacks are created with a 5s callback timeout, but serve runtime callback persistence was re-wrapping those contexts with `context.WithoutCancel(..., 10s)`.
- Added tests:
  - `TestServeRuntimePersistSnapshotPreservesParentDeadline`
  - `TestServeRuntimeAppendMessagesPreservesParentDeadline`
  - `TestServeRuntimeSnapshotCallbackPreservesEngineDeadline`
- Ran:
  - `gofmt -w cmd/serve_runtime.go cmd/serve_runtime_test.go`
  - `go build ./...`
  - `go test ./...`
  - `git diff --stat`
  - `git diff --check`

### DIFF
--- a/cmd/serve_runtime.go
+++ b/cmd/serve_runtime.go
@@ -449,14 +449,18 @@ func (rt *serveRuntime) ensurePersistedSession(ctx context.Context, sessionID st
 	return true
 }
 
+func inlinePersistContext(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
+	if ctx == nil {
+		ctx = context.Background()
+	}
+	return context.WithTimeout(ctx, timeout)
+}
+
 func (rt *serveRuntime) persistSnapshot(ctx context.Context, sessionID string, snapshot []llm.Message) bool {
 	if rt.store == nil || sessionID == "" {
 		return false
 	}
-	// Use a cancel-proof context so snapshot persistence succeeds even when the
-	// run context is cancelled (e.g. ^C or client disconnect), while preserving
-	// any context values (tracing, logging) from the original context.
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #839, opened 2026-06-19
fix: Completion notifications block the job worker for up to 10s after every run

sam-saffron-jarvis:
## What changed

- moved completion notifications off the worker hot path by making `finishRun()` enqueue notification work to an async goroutine after the terminal run state and final event are committed
- reused the already-loaded terminal `run` snapshot for notifications and loaded the job inside the async path
- kept retry scheduling independent of notification delivery
- updated notification tests to wait for the now-async side effects and added a regression test proving `finishRun()` returns quickly even when the notifier is blocked

## Why this is high-value

Before this change, every worker stayed occupied until `notifyRunDone()` finished, and that path can spend up to 10 seconds inside `notifyDone()` waiting on external delivery. That directly reduces queue throughput, delays retries, and makes low-worker setups feel stuck even though the run is already terminal in SQLite.

After this change, the worker returns to claiming the next queued run as soon as terminal DB state is committed. The remaining work is just an async notification goroutine, so notification latency no longer serializes job execution.

## Validation

- `gofmt -w cmd/serve_jobs_v2.go cmd/serve_jobs_v2_notify_test.go`
- `go build ./...`
- `go test ./...`
- added regression test: `TestJobsV2FinishRunDoesNotBlockOnCompletionNotification`
- updated existing notification tests to verify async delivery and async `notify_failed` event recording

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -1447,7 +1447,7 @@ func (m *jobsV2Manager) finishRun(runID string, status jobsV2RunStatus, result j
 		"input_tokens":  result.InputTokens,
 		"output_tokens": result.OutputTokens,
 	})
-	m.notifyRunDone(runID, status, result, exitReason, truncated, errText)
+	m.notifyRunDoneAsync(run, status, result, exitReason, truncated, errText)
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #840, opened 2026-06-19
fix: Finished jobs can remain stuck in claimed/running forever because finishRun errors are dropped

sam-saffron-jarvis:
## What changed

- Added a small `finishRunWithRetry` helper in `cmd/serve_jobs_v2.go` and routed every terminal `executeRun` path through it instead of discarding `finishRun` errors.
- The helper retries terminal status persistence up to 3 times with a short delay and logs loudly if finalization still cannot be persisted.
- Added a regression test that holds a transient SQLite write lock while a run is finishing, then verifies the run still reaches `succeeded` and no longer counts as active.

## Why this is high-value

A transient DB error during `finishRun` was previously dropped, leaving a run stuck in `claimed` or `running` until process restart. Because active-run counting includes those states, one bad finalization could block future runs for concurrency-limited jobs indefinitely, while also breaking retries/notifications and showing stale running jobs in the UI. Retrying the terminal update before `executeRun` returns closes that reliability hole for steady-state transient failures.

## Validation

- `go build ./...`
- `go test ./...`
- Added `TestJobsV2ExecuteRunRetriesFinishRunAfterTransientDBLock` to reproduce the transient-finalization-failure path and verify the run is not left active.

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log"
 	"net/http"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #847, opened 2026-06-20
fix: Reverse-hub request IDs can collide and corrupt concurrent in-flight requests

sam-saffron-jarvis:
## What changed

- Replaced reverse-hub request ID generation in `cmd/serve_hub_reverse.go` from `time.Now().UnixNano()` to a per-connection atomic counter via `hubReverseConnection.nextRequestID()`.
- Added `TestHubReverseConnectionNextRequestIDIsUnique` to validate that concurrent callers on the same reverse connection do not get duplicate request IDs.

## Why this is high-value

The reverse-hub protocol routes every response frame only by request ID. Before this change, two concurrent requests on the same reverse connection could generate the same timestamp-based ID and collide in the `pending` map. That could overwrite an in-flight waiter, misroute frames, or leave a caller hanging until cancellation. Using an atomic sequence makes request IDs collision-free within a connection, which is exactly the scope the protocol needs for safe routing.

## Validation

- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/cmd/serve_hub_reverse.go
+++ b/cmd/serve_hub_reverse.go
@@ -9,6 +9,7 @@ import (
 	"log"
 	"net/http"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/gorilla/websocket"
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #848, opened 2026-06-20
fix: File-backed session SQLite store still uses the default unbounded connection pool

sam-saffron-jarvis:
## What changed

- bound the session SQLite store to a single pooled connection for all databases, not just `:memory:`
- kept the existing `:memory:` behavior while also preventing file-backed stores from using `database/sql`'s default unbounded open-connection pool
- added a regression test covering file-backed stores so the pool bound stays in place

## Why this is high-value

The session store is on hot persistence paths for streamed responses, session resume, and serve/Telegram state updates. With SQLite in WAL mode, letting `database/sql` open many concurrent connections to the same file increases writer contention and pushes callers into `retryOnBusy()` loops. Serializing access through one shared pooled connection prevents that failure mode and reduces `SQLITE_BUSY` spikes without changing higher-level store behavior.

## Validation

- `go build ./...`
- `go test ./...`
- `go test -count=1 ./internal/session -run 'TestNewSQLiteStore(MemoryDBUsesSingleConnection|FileDBUsesSingleConnection)$' -v`
- `git diff --check`

### DIFF
--- a/internal/session/sqlite.go
+++ b/internal/session/sqlite.go
@@ -195,11 +195,13 @@ func NewSQLiteStore(cfg Config) (*SQLiteStore, error) {
 	if err != nil {
 		return nil, fmt.Errorf("open database: %w", err)
 	}
-	if dbPath == ":memory:" {
-		// Required for raw :memory: databases: keep a single connection so
-		// schema and session data stay visible to every operation.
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #852, opened 2026-06-21
fix: Reverse hub proxy aborts valid streams when the downstream consumer is merely slower than the upstream node

sam-saffron-jarvis:
## What changed

- Changed `hubReversePending.enqueue` in `cmd/serve_hub_reverse.go` to block until a response frame can be delivered or the request is canceled, instead of dropping frames when the fixed 16-frame buffer is full.
- Added coverage that a reverse-streamed response is backpressured rather than canceled when the downstream reader temporarily stops consuming.
- Updated the canceled-undrained-stream test to exercise the new blocked-enqueue path and verify cancellation still unblocks the websocket reader so later requests can proceed.

## Why this is high-value

Reverse-connected nodes are on the hot path for proxying and streaming responses. Before this change, a healthy stream was aborted as soon as the hub got roughly 16 queued `response_body` frames ahead of the downstream consumer, which is only about 512 KiB at the current 32 KiB chunk size. That makes slow browsers, WAN clients, mobile links, and large downloads fail spuriously even though the node and hub are both healthy.

This change turns that failure mode into normal backpressure: the hub stops reading more response frames for that request until the downstream side catches up or the request is canceled.

## Validation

- `go build ./...`
- `go test ./cmd -run 'TestHubReverse'`
- `go test ./...` *(fails on pre-existing unrelated test: `TestJobsV2ListJobsIncludesLastRunSummary` in `cmd/serve_jobs_v2_test.go`, with `LastRun = nil, want latest run summary`)*
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/cmd/serve_hub_reverse.go
+++ b/cmd/serve_hub_reverse.go
@@ -142,16 +142,9 @@ func (p *hubReversePending) enqueue(resp hubReverseResponse) bool {
 		return false
 	}
 	select {
-	case <-p.done:
-		return false
-	default:
-	}
-	select {
 	case p.ch <- resp:
 		return true
 	case <-p.done:
-		return false
-	default:
 		return false
 	}
 }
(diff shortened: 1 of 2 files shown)


SamSaffron:
Merged the high-impact reverse proxy backpressure fix into the working tree for inclusion via a combined local change.

PR #853, opened 2026-06-21
fix: Delegation ledger rewrites are non-atomic and can brick hub delegation state after a crash

sam-saffron-jarvis:
## What changed

- replaced direct `os.WriteFile` ledger rewrites in `internal/hub/delegation_store.go` with an atomic temp-file write in the same directory
- fsync the temp file before rename, rename it over the live ledger, re-assert `0600` permissions, and fsync the directory on supported platforms
- kept the existing JSON ledger format unchanged
- added a regression test that makes the ledger directory non-writable and verifies a failed update leaves the existing ledger readable, unchanged, and without leaked temp files

## Why this is high-value

The delegation ledger is used by active hub delegation flows. Before this change, every add/update truncated and rewrote the live JSON file in place. A crash, kill, or partial write at the wrong moment could leave invalid JSON on disk, after which later reads would fail hard and the hub delegation endpoints could return 500s until someone manually repaired the file.

This change removes that corruption window for the existing ledger contents: failed replacement attempts now leave the previously committed ledger intact and parseable.

## Validation

- `gofmt -w internal/hub/delegation_store.go internal/hub/delegation_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/hub/delegation_store.go
+++ b/internal/hub/delegation_store.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sort"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


SamSaffron:
Merged the high-impact atomic delegation ledger fix into the working tree for inclusion via a combined local change.

PR #854, opened 2026-06-21
fix: Local hub node store rewrites are non-atomic, so a partial write can make UI-added nodes disappear and fail auth

sam-saffron-jarvis:
## What changed

- Switched `internal/hub/store.go` from direct `os.WriteFile(...)` rewrites to the existing atomic temp-file + fsync + rename helper used for other critical persisted files.
- Kept the existing `0600` bearer-token protection behavior by tightening any existing store file before the atomic rewrite.
- Added a regression test that forces the atomic temp-file creation path to fail and verifies the previously persisted node store remains intact and readable.

## Why this is high-value

The local hub node store is the persistence layer for UI-added nodes and their bearer tokens. Before this change, add/upsert/remove rewrote the full JSON file in place. If the process crashed or the write was interrupted after truncation but before the full JSON landed, the store could become partial/invalid JSON.

Because `readLocked()` hard-fails on JSON parse errors and `Registry.Lookup()` ignores resolver errors, that corruption could make locally stored nodes silently disappear from hub lookups and break token-based auth for those nodes until manual repair. This change makes each rewrite atomic so the store is either fully updated or the previous valid file remains in place.

## Validation

- `gofmt -w internal/hub/store.go internal/hub/store_test.go`
- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/hub/store.go
+++ b/internal/hub/store.go
@@ -6,6 +6,8 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+
+	"github.com/samsaffron/term-llm/internal/config"
 )
 
 // Store persists UI-added nodes in a local JSON file (0600 — it holds node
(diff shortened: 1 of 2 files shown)


SamSaffron:
Merged the high-impact atomic hub node store fix into the working tree for inclusion via a combined local change.

PR #858, opened 2026-06-23
fix: `gitShowIndex` reads entire tracked blobs into memory before enforcing size caps

sam-saffron-jarvis:
## What changed

- Reworked `gitShowIndex` in `internal/tools/shell_track.go` to stream `git show :<path>` output through `StdoutPipe` and `io.LimitReader(maxBytes+1)` instead of calling `cmd.Output()`.
- If the tracked blob exceeds the per-file cap, the code now cancels the git command and returns `ok=false` before buffering the entire blob in memory.
- Added a shell tracking test covering the git-fallback path for a clean tracked file whose contents exceed `maxFileBytes`, ensuring the change still degrades to metadata-only recording.

## Why this is high-value

`buildChangeRecord` uses `gitShowIndex` to recover pre-edit contents for clean tracked files. Before this change, that path bypassed the surrounding snapshot protections by materializing the full blob in memory first and only then checking `len(out) > maxBytes`.

That meant a shell command touching a very large tracked file could cause a large one-shot allocation, extra copy work, and latency spikes in a long-lived Jarvis process, despite the file-tracking code otherwise being careful about per-file and per-call budgets. Streaming only up to `maxBytes+1` keeps the existing behavior while capping work and memory to the configured limit.

## Validation

- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/tools/file_track_test.go
+++ b/internal/tools/file_track_test.go
@@ -351,6 +351,50 @@ func TestShellToolGitFallback(t *testing.T) {
 	}
 }
 
+func TestShellToolGitFallbackRespectsMaxFileBytes(t *testing.T) {
+	dir := t.TempDir()
+	runGit := func(args ...string) {
+		t.Helper()
+		cmd := exec.Command("git", append([]string{"-C", dir}, args...)...)
+		cmd.Env = append(os.Environ(),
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented in the working tree: gitShowIndex now streams git show through a capped reader and oversized clean tracked files degrade to metadata-only. Verified with gofmt, go build ./..., and go test ./....

PR #859, opened 2026-06-23
fix: Shell file tracking runs full-repo `git status` twice on every shell tool call

sam-saffron-jarvis:
## What changed

- Skip the repo-wide `git status --porcelain -z --untracked-files=all` fallback when shell tracking already has narrower hints from `affected_paths` or session-tracked paths.
- Keep the existing full-repo git-status fallback for the no-hints case, so shell calls without any narrower candidate set still detect clean tracked, dirty tracked, and untracked repo changes.
- Added a regression test covering the bounded-tracking behavior when `affected_paths` is present in a git repo, while keeping the existing no-hints git fallback test.

## Why this is high-value

`shell` is a very hot tool path, and before this change every invocation inside a git repo paid for two whole-repo status scans: once before `cmd.Run` and once after. In large repos that work can dominate end-to-end tool latency.

This change removes both full-repo scans for the common case where the tool already has a bounded candidate set from `affected_paths` or prior session tracking. That cuts unnecessary subprocess work, repo traversal, and output parsing on every shell call, while preserving the broader best-effort fallback for shell commands that provide no narrower hints at all.

## Validation

- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/tools/file_track_test.go
+++ b/internal/tools/file_track_test.go
@@ -351,6 +351,57 @@ func TestShellToolGitFallback(t *testing.T) {
 	}
 }
 
+func TestShellToolGitAffectedPathsBoundTracking(t *testing.T) {
+	dir := t.TempDir()
+	runGit := func(args ...string) {
+		t.Helper()
+		cmd := exec.Command("git", append([]string{"-C", dir}, args...)...)
+		cmd.Env = append(os.Environ(),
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented in the working tree: shell tracking skips repo-wide git status when bounded affected_paths/session path hints are available while preserving the no-hints fallback. Verified with gofmt, go build ./..., and go test ./....

PR #860, opened 2026-06-23
fix: Response-run SSE subscribers are hard-dropped on brief backpressure, truncating live web streams

sam-saffron-jarvis:
## What changed

- Kept the existing overflow/drop behavior in `storeEventLocked`, but stopped treating it as a silent EOF on the SSE endpoint.
- Added `droppedSubscriberTerminalPayload` so an overflowed subscriber now gets a final `response.stream_error` SSE event with:
  - `error.type = stream_buffer_overflow`
  - current `sequence_number`
  - `min_replay_after`
  - `recovery` payload for resuming cleanly
- Updated `streamResponseRunEvents` to emit that terminal event and then write `[DONE]` instead of returning abruptly when a subscriber channel was closed due to overflow.
- Replaced the old overflow test expectation with a test that verifies the stream now emits both the structured terminal error and `[DONE]`.

## Why this is high-value

Today, a brief slow-client or proxy hiccup can fill the fixed per-subscriber buffer, close the live SSE stream, and make the browser/API client see an abrupt truncation even though the response run continues. That is directly user-visible and looks like flaky streaming.

This change is intentionally small and low-risk: it does not alter run execution or replay storage, and it does not change the existing overflow decision in the broadcaster. It fixes the reliability problem at the user boundary by turning a silent mid-stream death into an explicit recoverable terminal event that tells the client how to resume.

## Validation

- `go build ./...`
- `go test ./...`
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/cmd/serve_response_runs.go
+++ b/cmd/serve_response_runs.go
@@ -569,14 +569,24 @@ func (r *responseRun) subscribe(after int64) responseRunSubscribeResult {
 	return responseRunSubscribeResult{id: id, replay: replay, ch: ch}
 }
 
-func (r *responseRun) subscriberWasDropped(id int) bool {
+func (r *responseRun) droppedSubscriberTerminalPayload(id int) (map[string]any, bool) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
 	if !r.subscriberDropped[id] {
-		return false
+		return nil, false
 	}
 	delete(r.subscriberDropped, id)
-	return true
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented in the working tree: overflowed response-run SSE subscribers now receive response.stream_error with recovery data followed by [DONE], and the web client treats that terminal marker as resumable. Verified with gofmt, go build ./..., and go test ./....

PR #872, opened 2026-06-26
fix: Response completion does an uncancellable session-store lookup and can leave runs stuck in-progress forever

sam-saffron-jarvis:
## What changed

- Replaced both response-completion durable-response lookups with a bounded best-effort helper that uses `context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)`.
- Kept existing behavior of preferring the durable response id when available, but now completion falls back to the ephemeral `resp_*` id if the session-store lookup errors or times out.
- Added regression tests covering:
  - best-effort lookup still working even when the parent context is already canceled
  - synchronous completed-response storage no longer blocking forever on a stuck durable lookup

## Why this is high-value

Before this change, both completion paths used `context.Background()` for the session-store read that discovers the durable response id. If SQLite or the session store blocked there, the model work could already be finished but the run would never reach `response.completed`, never clear the active-run/session-busy state, never emit terminal `[DONE]`, and could delay shutdown waiting on the stuck goroutine.

This fix bounds that final store read to 5 seconds and makes it best-effort, so completion and cleanup still happen even when the durable lookup is unhealthy.

## Validation

- `gofmt -w cmd/serve_responses_durable.go cmd/serve_response_runs.go cmd/serve_response_runs_test.go`
- `go build ./...`
- `go test ./...`
- Added targeted regression tests in `cmd/serve_response_runs_test.go`

### DIFF
--- a/cmd/serve_response_runs.go
+++ b/cmd/serve_response_runs.go
@@ -1493,7 +1493,7 @@ func (s *serveServer) storeCompletedResponseRun(runtime *serveRuntime, sessionID
 			return "", err
 		}
 	}
-	durableID := s.latestDurableResponseIDForSession(context.Background(), sessionID)
+	durableID := s.latestDurableResponseIDForSessionBestEffort(context.Background(), sessionID)
 	completedID := respID
 	if durableID != "" {
 		completedID = durableID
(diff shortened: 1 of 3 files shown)


SamSaffron:
Implemented in local copy: both response completion paths now use bounded best-effort durable-id lookup instead of unbounded background lookup. Verified with go build ./... and targeted tests.

PR #876, opened 2026-06-28
fix: agentic ask drops assistant turns when response persistence fails once

sam-saffron-jarvis:
## What changed

- Return the `store.AddMessage` error from the `term-llm ask` response-completed persistence callback instead of reporting success after a failed assistant-message write.
- Added a focused regression test with a fake store that fails the first assistant `AddMessage`: the engine now treats the response callback as unhandled and redelivers the assistant through `TurnCompletedCallback` alongside the tool result.

## Why this is high-value

Agentic `ask` persists the assistant message before tool execution so sessions survive tool failures. When that write failed transiently, the callback still returned nil, causing the engine to omit the assistant from the later turn callback. That silently corrupted saved sessions by keeping tool results without the assistant/tool-call message that produced them.

Returning the write error preserves the existing engine recovery contract: a failed response callback means the assistant is included in the turn-completed messages and can be persisted on the retry path.

## Validation

- `go test ./cmd -run TestAskPersistenceRetriesAssistantThroughTurnCallbackWhenResponsePersistenceFails -count=1`
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/cmd/ask.go
+++ b/cmd/ask.go
@@ -148,6 +148,15 @@ func init() {
 	rootCmd.AddCommand(askCmd)
 }
 
+func persistAskAssistantResponse(ctx context.Context, store session.Store, sess *session.Session, assistantMsg llm.Message, durationMs int64, reasoningCfg config.ReasoningConfig) error {
+	sessionMsg := session.NewMessageWithReasoningPolicy(sess.ID, assistantMsg, -1, reasoningCfg)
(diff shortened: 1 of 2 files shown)


sam-saffron-jarvis:
Closing as duplicate of #877. The nightly parser has been updated to semantic-dedupe same-root-cause candidates before launching fix agents, including different titles/wrappers around the same failing callback contract.

PR #879, opened 2026-06-28
fix: jobs-v2 program runs can leave background children running after a successful command exits

sam-saffron-jarvis:
## What changed

- Jobs-v2 `program` runs now use the same process cleanup helper as the shell tool, so cleanup always best-effort kills the command process group and nonce-tagged descendants after the command returns.
- Restored the jobs-v2 program runner's existing `WaitDelay` after installing the shared cleanup helper, preserving the runner's prior wait-delay behavior.
- Added a regression test for a successful shell command that backgrounds a long-lived child, writes its PID, and exits immediately; the test asserts the child is gone after `Run` returns.

## Why this is high-value

Jobs-v2 is a daily automation path for Jarvis. Before this change, a successful `program` command could launch a background child and exit cleanly, causing Jarvis to mark the job successful while the child kept running indefinitely. Repeated scheduled runs could accumulate orphaned CPU/memory/file-handle consumers and make later operations harder to reason about.

## Validation

- Reproduced the failure with the new regression test before the fix: the background `sleep` process remained alive after `Run` returned.
- `gofmt -w cmd/serve_jobs_v2.go cmd/serve_jobs_v2_program_test.go`
- `go build ./...`
- `go test ./...`
- `go test ./cmd -run TestJobsV2ProgramRunner_CleansUpBackgroundChildOnSuccess -count=5`
- `git diff --check`

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -265,7 +265,6 @@ func (r *jobsV2ProgramRunner) Run(ctx context.Context, job jobsV2Job, pw progres
 	} else {
 		cmd = exec.CommandContext(ctx, cfg.Command, cfg.Args...)
 	}
-	cmd.WaitDelay = jobsV2ProgramWaitDelay
 	if strings.TrimSpace(cfg.Cwd) != "" {
 		cmd.Dir = cfg.Cwd
 	}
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing after manually bringing the validated changes into the working copy.

PR #880, opened 2026-06-29
fix: Serve/web fallback snapshot after compaction can delete pre-compaction scrollback

sam-saffron-jarvis:
## What changed

- Track when serve/web runtime history has become active-context-only after engine compaction.
- Route fallback/final snapshot reconciliation in that state through a compaction-aware SQLite replacement path that only rewrites rows at/after the current compaction boundary.
- Preserve pre-compaction scrollback rows and the compaction boundary while still reconciling post-boundary produced output.
- Add an integration-style serve runtime test that compacts, then hits a stream error that forces fallback snapshot persistence.

## Why this is high-value

Long-lived serve/web sessions can compact during ordinary daily use. Before this change, a later fallback full snapshot could pass only the compacted active context to `ReplaceMessages`, which treats its argument as the complete transcript, deletes pre-compaction rows, and clears the boundary. That could permanently remove persisted scrollback/search/audit history after a transient stream/tool/provider error.

## Validation

- `go test ./cmd -run 'TestServeRuntimeCompaction'`
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/cmd/serve_runtime.go
+++ b/cmd/serve_runtime.go
@@ -554,6 +554,48 @@ func (rt *serveRuntime) persistSnapshot(ctx context.Context, sessionID string, s
 	return true
 }
 
+type compactedMessageReplacer interface {
+	ReplaceCompactedMessages(ctx context.Context, sessionID string, messages []session.Message) error
+}
+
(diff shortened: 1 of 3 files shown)


SamSaffron:
Closing after manually bringing the validated changes into the working copy.

PR #884, opened 2026-07-01
FIX: per-session handover plan file, named upfront from the first user message

sam-saffron-jarvis:
## What

Planner → developer handovers were unreliable: the planner (especially on fable) would end up with its plan somewhere other than the file the handover mechanism reads. This PR makes the plan file per-session and authoritative, gives it a descriptive name upfront, and keeps concurrent planning per project intact.

### Per-session plan file, resolved from the system prompt

- `{{handover_path}}` still generates a unique random name per session, so concurrent sessions in one project never share a file. New `session.ExtractHandoverPath` recovers that exact path from the persisted system prompt — which survives resume, since the stored system message is reused rather than re-expanded.
- `/handover` file mode reads that exact file instead of scanning for the newest `.md`, so a concurrent session's plan, a spawned sub-agent's doc, or any stray file can't shadow it. The latest-scan survives only as a fallback when the prompt names no handover file.

### Descriptive filename from the first user message

- On the first user message of a handover-enabled session, the fast model produces two words describing the task (e.g. `auth-refactor`) and the plan file is named `2026-07-02-auth-refactor.md` immediately — set in stone upfront, instead of the old "wait for 1000 bytes then rename from content" flow (which is kept only as a fallback when no fast provider was available).
- The random path baked into the system prompt keeps working via a symlink; when the agent hasn't written yet, the symlink dangles until the first write creates the descriptive file. Name collisions between concurrent sessions keep one random word for uniqueness (`2026-07-02-auth-refactor-amber.md`).

### write_file/edit_file now write through symlinks

The atomic temp+rename in `write_file` and `edit_file` used to **replace** a symlink with a regular file. That silently broke the existing rename-with-symlink flow too (the agent's next write to the old path clobbered the link and content diverged). Both tools now resolve the link (including dangling ones) and rename onto its target.

### No permission prompts for the plan file

The plan lives under term-llm's own data dir (`~/.local/share/term-llm/handover/…`), outside the project's granted write roots, so every plan update triggered an approval prompt. `IsPathAllowedForRead/Write` now always allow paths inside the handover data root — it is app-managed state, not user files. Everything else in the data dir (e.g. `sessions.db`) stays protected.

### Planner prompt hardening

`system.md` states the path is decided upfront and fixed for the session, and forbids choosing a different filename, writing the plan into the repository/working directory, or creating extra `.md` files in the handover directory.

## Notes

- Tests: `ExtractHandoverPath` matching, `PrettifyHandoverName` (dangling-symlink naming, existing-content carry-over, collision suffix, two-word cap, non-random skip), write-through-symlink for the write tool, and a regression asserting `/handover` prefers the session's own plan over a newer stray `.md`. Full `go test ./...` passes (53 packages).

### DIFF
--- a/internal/agents/builtin/planner/system.md
+++ b/internal/agents/builtin/planner/system.md
@@ -11,7 +11,16 @@ Today is {{date}}. Working in {{cwd}}.
 
 ## Handover Document
 
-Maintain your plan in `{{handover_path}}`. Write to it incrementally as you work — update sections as your understanding evolves rather than writing everything at once. When the user runs `/handover @developer`, this file becomes the context for the next agent.
+Your plan lives at exactly this path, decided upfront and fixed for this session:
+
+`{{handover_path}}`
+
+Write to it incrementally as you work — update sections as your understanding evolves rather than writing everything at once. When the user runs `/handover @developer`, this exact file becomes the context for the next agent.
+
+Rules for this file — no exceptions:
+- Never choose a different filename or directory for the plan; the handover mechanism reads this exact path and nothing else
+- Never write the plan into the repository, the working directory, or any other location
+- Do not create additional .md files in the handover directory
 
 Structure your handover document with these sections:
 - **Objective** — what the user is trying to accomplish
--- a/internal/session/handover_path_test.go
+++ b/internal/session/handover_path_test.go
@@ -1,0 +1,49 @@
+package session
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+)
+
+func TestGetHandoverPathPinnedPerDirAndDate(t *testing.T) {
+	t.Setenv("XDG_DATA_HOME", t.TempDir())
+	ResetHandoverPathCache()
+	t.Cleanup(ResetHandoverPathCache)
+
+	first, err := GetHandoverPath(".", "2026-07-02")
+	if err != nil {
+		t.Fatalf("GetHandoverPath: %v", err)
+	}
+	base := filepath.Base(first)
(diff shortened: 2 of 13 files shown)


sam-saffron-jarvis:
Ran an external review of this branch with `chatgpt:gpt-5.5-high`. Findings and resolutions:

**Blocker — approval bypass via dangling symlink (fixed).** Permission checks canonicalize paths before approval, but a *dangling* link only resolves its parent; `resolveWriteTarget` then followed the link after approval. A dangling symlink planted in the always-writable handover dir pointing at e.g. `/tmp/x` would let a write escape approved roots without a prompt. Fix: `resolveWriteTarget` now only follows links whose target is a plain sibling name in the same directory — exactly what handover renames need — so cross-directory redirection is impossible by construction. Regression tests cover absolute, subdirectory, and `../` targets, plus an end-to-end write-tool escape attempt.

**Should-fix — fallback could still read another session's plan (fixed).** When the prompt named a pinned path but the file was empty/missing/dangling, `/handover` fell back to the newest `.md`, which could be a concurrent session's plan. Now the latest-scan runs only when the prompt names no handover path at all; an unwritten pinned plan means "no handover document yet" rather than "use someone else's". Regression test added.

**Should-fix — symlink tests on Windows (fixed).** All symlink-dependent tests skip via a probe helper when symlink creation is unsupported/unprivileged.

Declined: stricter `.md` boundary in `ExtractHandoverPath` — the prompt format is app-controlled and RE2 has no lookahead; left as-is.

PR #885, opened 2026-07-01
feat: smoother, friendlier web UI diff panel

sam-saffron-jarvis:
## What

Ten UX improvements to the web UI diff sidebar, plus two debug-provider fixes discovered while verifying them end to end.

**Diff panel (`app-diffs.js`, `app.css`, `index.html`, `app-core.js`):**

1. **Keyed incremental rendering** — file blocks are cached per path and patched in place; the list container is only touched when membership/order changes. No more full rebuild per event: text selection, focus, and scroll position survive live updates.
2. **Error state with retry** — a failed diff fetch used to leave "Loading diff…" forever; it now shows "Couldn't load this diff — Retry".
3. **Calmer live-follow** — only the file currently being edited stays auto-expanded (the previous one collapses unless the user pinned it open), with new expand-all/collapse-all header buttons and a subtle pulse on rows that update.
4. **Word-level diff emphasis** — adjacent del/add lines are paired and the changed span is marked, GitHub-style.
5. **Recency ordering** — most recently changed file sorts to the top instead of alphabetical.
6. **Chunked show-more** — huge diffs reveal 400 rows at a time (plus a "show all" escape hatch) instead of one synchronous render of thousands of rows; highlighting now gates on rendered rows, not total rows.
7. **Per-file copy actions** — hover a header for copy-path and copy-diff (reconstructs a unified patch from the cached hunks).
8. **Path filter** — appears automatically once the list reaches 8 files.
9. **Keyboard support** — Escape closes the drawer, arrow keys walk file headers, Enter/Space toggles, double-click on the resize handle resets the stored width.
10. **Badge pulse** — the toggle badge pulses when totals change so activity is discoverable while the panel is closed.

**Debug provider fixes (found via the e2e below):**
- Tool commands sent the stale `file_path` argument; the tools expect `path`, so every `write`/`edit`/`read` DSL command failed with INVALID_PARAMS.
- Any tool result in history short-circuited to canned completion text, so DSL commands only worked on a session's first turn. Now only a trailing tool round short-circuits.

## Testing

- `app_diffs_test.js`: 35 cases pass (12 new covering emphasis, retry, chunking, filter, expand/collapse-all, Escape, resize reset, node-identity reuse, unified-diff builder, ordering).
- Full `go test ./...` passes.
- **Clean-room Playwright e2e**: built the binary, ran `serve web --no-auth --provider debug --yolo` under a throwaway `$HOME`/workspace, drove real runs through the composer (debug DSL emits real tool calls -> real file-tracking events), and verified all ten behaviors in headless Chromium — 16 checks passed, including fault-injected fetch failures (route abort) and drawer mode across the viewport breakpoint. Environment fully torn down afterwards.

### DIFF
--- a/internal/llm/debug_provider.go
+++ b/internal/llm/debug_provider.go
@@ -214,8 +214,11 @@ func (d *DebugProvider) Capabilities() Capabilities {
 // Otherwise, streams the debug markdown content.
 func (d *DebugProvider) Stream(ctx context.Context, req Request) (Stream, error) {
 	return newEventStream(ctx, func(ctx context.Context, send eventSender) error {
-		// If tool results are present, stream completion text
-		if hasToolResults(req.Messages) {
+		// If the request is completing a tool round (the conversation ends
+		// with tool results), stream completion text. Earlier turns' tool
+		// results must not short-circuit new user commands in multi-turn
+		// sessions.
+		if lastMessageHasToolResults(req.Messages) {
 			return d.streamCompletionText(ctx, send)
 		}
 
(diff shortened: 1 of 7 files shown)


sam-saffron-jarvis:
Second commit (`be1cdf90`) addresses an external model review (gpt-5.5-high) of the first commit. All four should-fix findings plus three nits:

- **Stale body under unchanged seq** — body rebuilds are now keyed on a per-write cache revision instead of the locally-known seq, and authoritative list refreshes mark cached diffs dirty when the server seq is ahead (covers tabs that missed events while detached).
- **Canonical path replacement** — all path-keyed state (expanded/collapsed/limits/caches/errors/blocks) is reconciled when the server list canonicalizes live paths, and live-follow remaps to the most recent entry instead of silently detaching.
- **Nested action buttons vs header keydown** — Enter/Space on the copy buttons no longer toggles the accordion.
- **Surrogate pairs** — emphasis prefix/suffix trimming now compares code points, so marks never split an emoji.
- Copy actions route through the app-wide `getClipboardWriter()` (execCommand fallback); transient pulse/updated/copied classes clear pending timers before re-triggering; `lastMessageHasToolResults` is pinned by three new direct Go tests.

Re-verified: 45 node unit tests (5 new), full `go test ./...`, and the clean-room Playwright e2e re-run — now 17 checks including a new one for the keydown guard — all green.

PR #886, opened 2026-07-02
fix: Uploaded input_file payloads are persisted as full base64 blobs in session rows

sam-saffron-jarvis:
## What changed

- Strip `PartFile.FileData.Base64` from persisted session parts when the file has already been saved under term-llm's uploads directory, while retaining the file path, media type, filename, and size metadata.
- Rehydrate path-only uploaded files for Responses native `input_file` payloads just before request construction, only for first-party upload paths and within the existing upload size cap/policy.
- Added coverage for parsed `input_file` data URLs, storage sanitization, and native Responses rehydration from a saved upload path.

## Why this is high-value

Large web/API file uploads were being stored twice: once as decoded bytes in the uploads directory and again as a ~33% larger base64 string in the SQLite `messages.parts` JSON. For near-limit files, that bloats session rows by tens of MiB and slows session writes, reads, backups, and resume/search I/O. The fix removes that duplicate serialized payload while preserving native file resend behavior by reading the already-saved upload only when a Responses request can use it.

## Validation

- `go test ./internal/session ./internal/llm ./cmd`
- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/cmd/serve_test.go
+++ b/cmd/serve_test.go
@@ -1246,6 +1246,18 @@ func TestParseResponsesInput_FileUploadSavesToDisk(t *testing.T) {
 	if strings.Contains(msg.Parts[0].Text, dataHome) || strings.Contains(msg.Parts[0].Text, entries[0].Name()) {
 		t.Fatalf("parts[0].text leaks upload path: %q", msg.Parts[0].Text)
 	}
+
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #894, opened 2026-07-06
fix: Telegram session creation holds the global session mutex while doing blocking DB writes

sam-saffron-jarvis:
## What changed

- Publish the winning Telegram session under `telegramSessionMgr.mu`, then release the global session mutex before writing session metadata to the store.
- Give session metadata persistence its own 5s `context.WithTimeout(context.WithoutCancel(ctx), ...)` so a canceled or unbounded serve context cannot leave `Create`/`SetCurrent` wedged forever.
- Add regression coverage for a blocked `Create` call: another chat can still create and return its session while the first persistence call is stalled.
- Add coverage that `persistSession` passes a bounded, non-canceled context to store writes.

## Why this is high-value

Sam uses the Telegram/Jarvis path daily. Before this change, a slow or locked session store write during new/reset session persistence held the single `telegramSessionMgr.mu`, blocking every other chat's `getOrCreate`, reset, and shutdown session-map access. Moving only the blocking I/O out of the mutex preserves the winner decision while preventing one SQLite/filesystem stall from making the Telegram bot globally appear stuck.

## Validation

- `go build ./...`
- `go test ./internal/serve -run 'TestTelegramSessionMgr(GetOrCreate_DoesNotBlockOtherChatsWhilePersistingSession|PersistSession_UsesBoundedContextDetachedFromCallerCancel|GetOrCreate_PersistsOnlyWinningConcurrentSession|ResetSessionIfCurrent_RejectsStaleExpectedSession)'`
- `git diff --check`
- `go test ./...` was run twice; both runs reached the changed `internal/serve` tests successfully but failed in unchanged `internal/jobs.TestProgramRunnerDoesNotLeakBackgroundChildren` because its background child process was still observed alive after 1s.

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -522,8 +522,9 @@ func (m *telegramSessionMgr) getOrCreate(ctx context.Context, chatID int64) (*te
 		return existing, nil
 	}
 	m.sessions[chatID] = created
-	m.persistSession(ctx, created)
 	m.mu.Unlock()
+
+	m.persistSession(ctx, created)
 	return created, nil
 }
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #898, opened 2026-07-09
fix: Telegram runner cleanup can hang forever after the stream watchdog fires

sam-saffron-jarvis:
## What changed

- Bound Telegram `settings.Runner` cleanup in `streamReply` so it no longer waits forever on `runnerDone` after cancellation.
- Cancel the stream before the cleanup wait for early-return paths, then release the Telegram session lock after the bounded timeout if the runner is still stuck.
- Clear active runner engine state on cleanup and guard runner `OnEngineReady`/`OnEngineDone` callbacks with a per-stream token so late callbacks from an abandoned runner cannot corrupt a later stream.
- Added a regression test with a runner that observes cancellation but does not return until explicitly released; the test asserts `streamReply` returns after the bounded cleanup timeout and a second message can run afterward.

## Why this is high-value

Telegram/Jarvis streaming uses a session mutex to serialize each chat. The stream watchdog already cancels stuck streams, but the runner cleanup path did an unconditional receive on `runnerDone` after logging the 5s timeout. If a provider/tool/background operation ignored cancellation, that defer could keep `sess.mu` held forever, wedging all future messages in that chat despite the watchdog firing.

This keeps normal cleanup behavior when the runner exits promptly, while preventing one stuck runner from permanently blocking Sam's daily Telegram session.

## Validation

- `go build ./...`
- `go test ./...`
- `git diff --check`

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -34,6 +34,10 @@ const minEditInterval = 3 * time.Second
 // defaultStreamEventTimeout is used when telegramSessionMgr.streamEventTimeout is zero.
 const defaultStreamEventTimeout = 10 * time.Minute
 
+// defaultRunnerCleanupTimeout bounds how long streamReply waits for the shared
+// runner to observe cancellation before releasing the Telegram session lock.
+const defaultRunnerCleanupTimeout = 5 * time.Second
+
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #899, opened 2026-07-09
fix: EventPipe runner adapters can hang forever after cancellation

sam-saffron-jarvis:
## What changed

- Added a shared `run.WaitForRunnerDone` helper with a bounded cleanup deadline for EventPipe runner adapters.
- Updated the TUI chat, `ask`, and Telegram EventPipe runner paths to cancel the run context and detach instead of waiting forever when `Runner.Run` ignores cancellation.
- Avoided the old `ask` shared-result data race risk by passing runner results through a buffered outcome channel and only applying them when the runner actually finishes.
- Marked Telegram sessions with a detached runner as stale so the next message creates a fresh runtime instead of reusing a possibly still-active shared engine.
- Updated/added tests covering the bounded wait helper and the TUI cancellation path where a fake runner never returns.

## Why this is high-value

These EventPipe paths sit on daily chat/ask/Telegram streaming. If a provider, tool, MCP cleanup, or runner path wedges after cancellation, the old code kept waiting on the runner forever after the UI/stream consumer had already stopped. That can freeze TUI stream cleanup, hang `ask` after output cancellation, and keep Telegram chats blocked behind the per-session mutex.

Bounding the post-cancel runner cleanup wait preserves the normal graceful path while preventing one stuck runner from wedging the caller indefinitely.

## Validation

- `gofmt -w internal/run/runner_wait.go internal/run/runner_wait_test.go internal/tui/chat/streaming.go internal/tui/chat/streaming_test.go cmd/ask.go internal/serve/telegram.go`
- `go test ./internal/run ./internal/tui/chat ./internal/serve ./cmd`
- `go build ./...`
- `git diff --check --cached`
- `go test ./...` was run; it fails in unchanged `internal/jobs` at `TestProgramRunnerDoesNotLeakBackgroundChildren` because a background child process still appears alive after the test deadline.

### DIFF
--- a/cmd/ask.go
+++ b/cmd/ask.go
@@ -70,6 +70,8 @@ var (
 	askSkills string
 	// Session resume flag
 	askResume string
+
+	askRunnerCleanupTimeout = runpkg.DefaultRunnerCleanupTimeout
 )
 
 var askCmd = &cobra.Command{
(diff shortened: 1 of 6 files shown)


SamSaffron:

PR #902, opened 2026-07-10
fix: Every shell call renders the full approval transcript before fast-path approval

sam-saffron-jarvis:
## What changed

- Added a lazy shell approval path that accepts a transcript supplier instead of an already-materialized transcript.
- Switched `ShellTool.Execute` to pass `shellApprovalTranscriptFromContext(ctx)` as that supplier, so transcript rendering is deferred until an auto guardian reviewer actually needs it.
- Kept the existing eager `CheckShellApprovalWithContext` API for callers that already build transcript evidence, such as handover approval.
- Added tests proving the supplier is not invoked for yolo or deterministic shell approvals, and is invoked exactly once for guardian review.

## Why this is high-value

Shell is one of Jarvis's hottest tool paths. Before this change, every shell invocation rendered the full approval transcript before `CheckShellApprovalWithContext` could take the yolo, configured allowlist, session cache, guardian exact cache, or project approval fast paths. Rendering copies the active message slice and materializes text/tool-call/tool-result evidence, so the wasted work grows with session length and repeats for every shell call.

With the transcript behind a supplier, fast-path-approved commands skip that copy/render work entirely. Only unresolved auto-mode commands with an available guardian reviewer pay the transcript cost, preserving guardian evidence while making common approved shell calls independent of transcript size.

## Validation

- `gofmt -w internal/tools/approval.go internal/tools/shell.go internal/tools/approval_auto_test.go`
- `go build ./...`
- `go test ./internal/tools -run 'TestApprovalManager(LazyTranscriptSupplier|AutoReviewerOnlyAfterDeterministicMiss|GuardianExactCacheDoesNotTreatStarAsPattern)'`
- `tmp=$(mktemp -d); mkdir -p "$tmp/home" "$tmp/xdg" "$tmp/codex"; HOME="$tmp/home" XDG_CONFIG_HOME="$tmp/xdg" CODEX_HOME="$tmp/codex" go test ./...` (isolated user skill dirs so the existing cmd skill-visibility test is not affected by Jarvis's ambient user skills)
- `git diff --check`

### DIFF
--- a/internal/tools/approval.go
+++ b/internal/tools/approval.go
@@ -972,11 +972,20 @@ func getDirectoryForApproval(path string) string {
 // CheckShellApproval checks if a shell command is approved.
 // workDir is the directory where the command will execute (may be empty for cwd).
 func (m *ApprovalManager) CheckShellApproval(command, workDir string) (ConfirmOutcome, error) {
-	return m.CheckShellApprovalWithContext(context.Background(), command, workDir, nil)
+	return m.checkShellApprovalWithContext(context.Background(), command, workDir, nil)
 }
 
 // CheckShellApprovalWithContext checks shell approval with optional transcript evidence.
+// It keeps the existing eager transcript API for callers that have already
+// materialized approval evidence.
 func (m *ApprovalManager) CheckShellApprovalWithContext(ctx context.Context, command, workDir string, transcript []TranscriptEntry) (ConfirmOutcome, error) {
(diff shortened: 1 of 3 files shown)


SamSaffron:
Implemented the still-needed lazy approval transcript fix directly in the current worktree, preserving the current structured guardian events. Superseded by the local changes.

PR #903, opened 2026-07-10
fix: WebSocket completion rebuilds and retains the full transcript after every turn

sam-saffron-jarvis:
## What changed

- Stop rebuilding the full Responses input after each successful WebSocket turn just to populate `wsLastRequest`.
- Store a lightweight metadata snapshot for future continuation compatibility checks, clearing `Input`, `Messages`, and `PreviousResponseID` before retaining it.
- Add a focused fake-WebSocket continuation test that asserts completion does not call `buildFullInput` and does not retain transcript slices in `wsLastRequest`.

## Why this is high-value

`wsLastRequest` is only used by `responsesRequestNonInputEqual`, which ignores transcript fields. Keeping full input there made every completed continuation turn pay transcript-sized conversion/copy work and retain duplicate history until the socket was discarded. The new path preserves the compatibility check while making completion work independent of accumulated transcript length and avoiding redundant transcript retention in long-lived Jarvis/OpenAI WebSocket sessions.

## Validation

- `gofmt -w internal/llm/responses_websocket.go internal/llm/responses_websocket_test.go`
- `go build ./...`
- `go test ./internal/llm -run 'TestResponsesWebSocket(PrepareUsesContinuationWithoutFullInputRebuild|CompletionStoresLightweightLastRequest)'`
- `tmp=$(mktemp -d); HOME="$tmp" XDG_CONFIG_HOME="$tmp/.config" GOMODCACHE=/home/agent/go/pkg/mod GOCACHE=/home/agent/.cache/go-build go test ./...`
- `git diff --stat && git diff --check`

### DIFF
--- a/internal/llm/responses_websocket.go
+++ b/internal/llm/responses_websocket.go
@@ -228,10 +228,13 @@ func (c *ResponsesClient) streamWebSocketPrepared(ctx context.Context, req Respo
 		if err := handler.Finish(send); err != nil {
 			return err
 		}
-		fullReq := wireReq
-		fullReq.Input = append([]ResponsesInputItem(nil), buildFullInput()...)
-		fullReq.PreviousResponseID = ""
-		c.wsLastRequest = &fullReq
+		lastReq := wireReq
+		// Future continuation checks only compare non-input request metadata. Do not
+		// rebuild or retain the full transcript after every completed WebSocket turn.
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented the still-needed lightweight WebSocket request snapshot fix directly in the current worktree. Superseded by the local changes.

PR #905, opened 2026-07-11
fix: unblock WebRTC peer cleanup on idle timeout

sam-saffron-jarvis:
## What

- close the underlying SCTP association whenever a WebRTC data channel is stopped
- use a narrow data-channel interface so shutdown behavior is regression-testable
- add a test modelling Pion's asymmetric close semantics: closing the data channel does not wake a blocked reader, while closing the transport does

## Why

The web peer's idle-timeout path called `DataChannel.Close()` and then waited on `readDone`. Pion's data-channel close resets only the SCTP stream's write direction, so a goroutine blocked in `ReadDataChannel` could remain blocked forever.

That also kept `handleOffer` alive and prevented the active connection counter from decrementing. On the affected deployment, pprof showed **8 leaked handlers plus 2 live handlers = the configured 10-connection maximum**. New WebRTC offers were then rejected and browsers had to fall back to HTTPS, creating intermittent connection delays.

Closing the SCTP association wakes the blocked stream reader before `runDataChannel` waits for it.

## Validation

- `go test ./internal/webrtc`
- `go test -race ./internal/webrtc`
- `go build ./...`
- `go test ./...` *(two unrelated failures on current main: a user-config-dependent `cmd` skill-visibility test and a flaky jobs child-process cleanup test; the jobs test passed when rerun alone)*
- `git diff --check`

### DIFF
--- a/internal/webrtc/peer.go
+++ b/internal/webrtc/peer.go
@@ -63,6 +63,12 @@ type responseFrame struct {
 	Status  int               `json:"status,omitempty"`  // HTTP status; for "headers" and "done"
 }
 
+type dataChannelConn interface {
+	ReadDataChannel([]byte) (int, bool, error)
+	WriteDataChannel([]byte, bool) (int, error)
+	Close() error
+}
+
 const maxChunkDataBytes = 16 * 1024
 
 // peer is the home-side WebRTC peer.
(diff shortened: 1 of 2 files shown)


SamSaffron:
Implemented the still-needed WebRTC cleanup fix directly in the current worktree. Superseded by the local changes.

PR #908, opened 2026-07-11
fix: Worktree listing performs 7N Git launches and N full session scans

sam-saffron-jarvis:
## What changed

- Batched worktree in-use session lookup for the web worktree list endpoint by listing session summaries once and grouping them by canonical worktree path.
- Reworked managed worktree listing to build metadata from `git worktree list --porcelain` instead of re-running per-worktree root/head/branch/base Git probes.
- Kept dirty-file checks as the only per-worktree Git operation and run those checks with bounded concurrency.
- Added regression coverage that verifies one session-list call for multiple worktrees and uses a counting Git wrapper to reject per-worktree metadata probes during `worktree.List`.

## Why this is high-value

The worktree selector is on the interactive web path and is refreshed on initial load, menu open, and after worktree actions. Previously it did one full session summary scan per worktree and around seven serial Git launches per row. This change reduces the session-store work from O(worktrees × sessions) to O(worktrees + sessions), and reduces per-row Git work to only dirty status checks.

## Validation

- `go test ./internal/worktree ./cmd -run 'TestListUsesPorcelainMetadataWithoutPerWorktreeGitProbes|TestServeWorktreeListBatchesSessionUsage' -count=1`
- `go build ./...`
- `git diff --check`
- `go test ./...` was run, but this checkout still fails on unrelated existing tests in this environment:
  - `cmd`: `TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` does not see the project-local test skill because Sam's user skill registry hits `max_visible_skills=8` first.
  - `internal/jobs`: `TestProgramRunnerDoesNotLeakBackgroundChildren` intermittently reports a still-live background child; rerunning that test alone passed during validation.

### DIFF
--- a/cmd/serve_worktrees.go
+++ b/cmd/serve_worktrees.go
@@ -146,9 +146,13 @@ func (s *serveServer) handleWorktreeList(w http.ResponseWriter, r *http.Request)
 		writeOpenAIError(w, http.StatusInternalServerError, "server_error", err.Error())
 		return
 	}
+	dirs := make([]string, 0, len(items))
 	for _, wt := range items {
-		inUse, _ := worktree.InUse(r.Context(), s.store, wt.Dir)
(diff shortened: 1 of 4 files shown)


SamSaffron:
Implemented the still-needed fix directly in the current worktree, adapted to preserve active-session-only cleanup semantics and canonical path matching. Superseded by the local changes.

PR #911, opened 2026-07-12
fix: Ripgrep can hang forever after stderr exceeds the 64 KiB capture limit

sam-saffron-jarvis:
## What changed

- Continue draining ripgrep's stderr pipe to EOF after retaining the existing bounded 64 KiB diagnostic prefix.
- Add a regression test that places a fake `rg` first in `PATH`, writes 4 MiB to stderr, and verifies `runRipgrep` returns promptly with bounded diagnostic text.

## Why this is high-value

The grep tool is a daily agent path. Previously, once ripgrep wrote more than the 64 KiB capture limit plus the OS pipe capacity, the stderr reader stopped consuming data and ripgrep blocked on its full pipe. The stdout reader then waited forever for EOF, wedging the active provider/tool loop. Draining excess diagnostics prevents that deadlock without increasing diagnostic memory use or changing successful grep behavior.

## Validation

- `go build ./...`
- `go test ./internal/tools -run TestRunRipgrepDrainsStderrBeyondCaptureLimit -count=10`
- `go test ./...` (all changed-package tests pass; the repository-wide run encountered unrelated existing failures in `cmd` skill-visibility isolation and `internal/jobs` background-child cleanup)
- `git diff --check`

### DIFF
--- a/internal/tools/grep.go
+++ b/internal/tools/grep.go
@@ -326,6 +326,7 @@ func runRipgrep(ctx context.Context, args []string, limits ripgrepCaptureLimits,
 	stderrCh := make(chan []byte, 1)
 	go func() {
 		data, _ := io.ReadAll(io.LimitReader(stderr, 64*1024))
+		_, _ = io.Copy(io.Discard, stderr)
 		stderrCh <- data
 	}()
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
Validated the bug and fix. Closing this PR because I'm applying the fix directly to the current main working tree, with safer pipe-wait ordering and success-path regression coverage added.

PR #913, opened 2026-07-13
fix: Telegram silently drops photos sent during an active run

sam-saffron-jarvis:
## What changed

- Force Telegram photo messages received during an active stream through the existing cancel-and-replace path instead of reducing them to text interjections.
- Preserve the full structured image message and its temporary backing path until the replacement provider turn consumes it.
- Add an integration test with a blocked active stream and a captionless photo, asserting that the photo cancels the old stream, reaches the replacement provider request as an image, and still has a valid local path during provider consumption.

## Why this is high-value

Captionless photos sent while Jarvis was busy were deterministically classified as empty text interjections. Telegram acknowledged them as “Noted,” but the engine received only an empty text message and the handler deleted the downloaded image. Screenshots and photos used to steer a long-running Telegram task were therefore silently lost. Treating attachments as replacement requests uses the already established safe lifecycle and avoids acknowledging work that Jarvis cannot see.

## Validation

- `go build ./...`
- `go test ./internal/serve -run '^TestHandleMessage_PhotoInterruptCancelsAndPreservesImage$' -count=10`
- `git diff --check`
- `go test ./...` was also run; the changed `internal/serve` package passed, while two unrelated existing/environment-sensitive tests failed: `cmd/TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` (host skill inventory truncation) and `internal/jobs/TestProgramRunnerDoesNotLeakBackgroundChildren` (background child still reported alive).

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -1366,8 +1366,14 @@ func (m *telegramSessionMgr) handleMessage(ctx context.Context, bot telegramBot,
 			}
 			sess.cancelMu.Unlock()
 
-			fastProvider := m.newFastProvider()
-			action := llm.ClassifyInterrupt(ctx, fastProvider, newMsgText, activity)
+			// Photo interjections cannot outlive this handler's temporary image path.
+			// Cancel and process them as replacement turns so the full structured
+			// message remains available until the provider consumes it.
+			action := llm.InterruptCancel
+			if tempImagePath == "" {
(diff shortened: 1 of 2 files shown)


SamSaffron:
Validated the bug and fix. Closing this PR because I'm applying the fix directly to the current main working tree, with stronger temporary-image lifecycle coverage added.

PR #914, opened 2026-07-13
fix: Parallel image generations overwrite each other

sam-saffron-jarvis:
## What changed

- Save generated images with `O_CREATE|O_EXCL` so an existing filename can never be truncated.
- Retry collisions with a numeric filename suffix while writing through the exclusively created file.
- Add a concurrent regression test that saves 32 distinct payloads for the same prompt and verifies every returned path and file contents are unique.

## Why this is high-value

Parallel tool calls can complete multiple paid image generations in the same second. The old seconds-resolution prompt-based filename made those calls overwrite one another while all callers reported success, permanently losing generated artifacts and leaving tool results or memory records pointing at the same final image. Exclusive creation preserves every output without serializing image generation.

## Validation

- `gofmt -w internal/image/output.go internal/image/output_test.go`
- `go build ./...`
- `go test -race ./internal/image -run '^TestSaveImageConcurrentCallsDoNotOverwrite$' -count=10`
- Full suite passed in an isolated skill environment with the unrelated baseline `TestProgramRunnerDoesNotLeakBackgroundChildren` test skipped: `go test ./... -skip '^TestProgramRunnerDoesNotLeakBackgroundChildren$'`
- A plain `go test ./...` was also run; it hit environment/baseline failures outside this change (`TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` because host-global skills exceed its eight-skill test limit, and, after isolating HOME, `TestProgramRunnerDoesNotLeakBackgroundChildren`).
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/internal/image/output.go
+++ b/internal/image/output.go
@@ -23,13 +23,36 @@ func SaveImage(data []byte, outputDir, prompt string) (string, error) {
 	}
 
 	filename := generateFilename(prompt)
-	path := filepath.Join(dir, filename)
+	ext := filepath.Ext(filename)
+	stem := strings.TrimSuffix(filename, ext)
 
-	if err := os.WriteFile(path, data, 0644); err != nil {
-		return "", fmt.Errorf("failed to write image: %w", err)
+	for suffix := 0; ; suffix++ {
+		candidate := filename
+		if suffix > 0 {
+			candidate = fmt.Sprintf("%s-%d%s", stem, suffix+1, ext)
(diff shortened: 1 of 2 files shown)


SamSaffron:
Validated the bug and fix. Closing this PR because I'm applying the fix directly to the current main working tree, with deterministic collision coverage and short-write handling added.

PR #918, opened 2026-07-14
fix: Job cancellation can report success while the newly started job keeps running

sam-saffron-jarvis:
## What changed

- Made queued/claimed and running cancellation updates conditional on the status that was observed.
- Added a bounded retry when another worker advances or completes the run before the cancellation update wins.
- Ensured a claimed-to-running race is re-read as running, transitioned to `cancel_requested`, and signalled through the registered cancel function.
- Added deterministic coverage for both claimed-to-running cancellation and running-to-succeeded completion races.

## Why this is high-value

Previously, cancelling immediately after triggering a job could overwrite a newly running row as `cancelled` without cancelling its runner. The API would report a terminal cancellation while the shell command or LLM agent continued consuming tokens, running tools, and producing filesystem or network side effects. The same stale snapshot could overwrite a just-completed terminal result with `cancel_requested`.

The compare-and-swap status transitions keep the persisted result and the actual runner lifecycle aligned in both races.

## Validation

- `go build ./...`
- `XDG_CONFIG_HOME=$(mktemp -d) go test ./cmd`
- `go test -race ./cmd -run 'TestJobsV2CancelRun(FollowsClaimedToRunningTransition|DoesNotOverrideConcurrentCompletion)$' -count=1`
- `XDG_CONFIG_HOME=$(mktemp -d) go test ./...` passes all packages except the unchanged `internal/jobs` test `TestProgramRunnerDoesNotLeakBackgroundChildren`, whose killed child remains visible on this host beyond its one-second polling deadline.
- `git diff --check`

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -530,14 +530,15 @@ type jobsV2Manager struct {
 	lastCleanupCompleted time.Time
 	cleanupRunning       bool
 
-	enqueueMu     sync.Mutex
-	mu            sync.Mutex
-	closed        bool
-	done          chan struct{}
-	schedulerWake chan struct{}
-	workerWake    chan struct{}
-	wg            sync.WaitGroup
-	cancels       map[string]context.CancelFunc
+	enqueueMu         sync.Mutex
+	mu                sync.Mutex
+	closed            bool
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #922, opened 2026-07-15
fix: Late image-recorder wiring drops metadata and leaks a SQLite connection per runtime

sam-saffron-jarvis:
## What changed

- Patch the already-registered `image_generate` tool when `SetImageRecorder` is called, including its agent and session attribution.
- Replace runtime-lifetime ownership of an eagerly opened memory store with a lazy recorder that opens the configured SQLite store only for `RecordImage` and closes it immediately afterward.
- Add an end-to-end regression test covering late recorder wiring, attribution, zero opens during setup, and one open/close pair for one generated image.

## Why this is high-value

Every tool-enabled runtime previously opened and retained a memory SQLite store without any lifecycle owner to close it. At the same time, registration happened before recorder wiring, so the existing `image_generate` instance kept a nil recorder and silently omitted generated-image metadata from memory. This fix restores image search/attribution for ordinary Jarvis image generation while preventing per-runtime database connections, goroutines, and file descriptors from accumulating in the long-running service.

## Validation

- `gofmt -w internal/tools/registry.go cmd/memory.go cmd/memory_image_recorder_test.go`
- `go build ./...`
- `go test ./cmd -run '^TestWireImageRecorderIsLazyAndRecordsWithAttribution$' -count=10`
- `go test ./internal/tools`
- `git diff --check`
- `go test ./...` was also run. All relevant tests passed; the suite still reports two unrelated existing/environment-sensitive failures: `TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` (the host's 30 configured skills exceed that test's 8-skill visibility cap) and `TestProgramRunnerDoesNotLeakBackgroundChildren` (its background child remains visible in this host environment).

### DIFF
--- a/cmd/memory.go
+++ b/cmd/memory.go
@@ -58,30 +58,71 @@ func init() {
 	memoryCmd.AddCommand(memoryInsightsCmd)
 }
 
-func openMemoryStore() (*memorydb.Store, error) {
-	if dbEnv := os.Getenv("TERM_LLM_MEMORY_DB"); dbEnv != "" {
-		if memoryDBPath == defaultMemoryDBPath {
-			memoryDBPath = dbEnv
-		}
+type imageRecordStore interface {
+	tools.ImageRecorder
+	Close() error
+}
+
+type imageRecordStoreOpener func(path string) (imageRecordStore, error)
+
+type lazyImageRecorder struct {
+	path string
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #923, opened 2026-07-15
fix: Tool-enabled runtimes leak a memory SQLite store and connection

sam-saffron-jarvis:
## What changed

- Replaced eager memory-store wiring with a lightweight `ImageRecorder` that opens the memory SQLite store only when `RecordImage` is called.
- Closed the store after every image record attempt, so no SQL connection remains owned by a tool manager.
- Added a focused regression test that wires 100 tool managers without creating the memory database, then verifies one image record causes exactly one open, one record, and one close.

## Why this is high-value

`SessionSettings.SetupToolManager` runs for ordinary tool-enabled web, Telegram, job, and subagent runtimes. Previously every setup initialized the memory schema and retained an unclosed `sql.DB`, even when image generation was never used. Repeated runtime creation could therefore accumulate SQLite connections, file descriptors, connection-opener goroutines, and memory in the long-running daemon.

The lazy recorder removes that per-runtime resource ownership without broad lifecycle changes and preserves best-effort image recording behavior.

## Validation

- `gofmt -w cmd/memory.go cmd/memory_recorder_test.go`
- `go build ./...`
- `go test ./cmd -run TestWireImageRecorderLazilyOpensAndClosesStore -count=1`
- `XDG_CONFIG_HOME=$(mktemp -d) go test ./cmd`
- `git diff --check`

`go test ./...` was also run. All relevant packages passed, but the repository-wide command encountered two unrelated existing/environmental failures: `TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` reads the host's eight-skill display cap (and passes with an isolated `XDG_CONFIG_HOME`), while `TestProgramRunnerDoesNotLeakBackgroundChildren` in unchanged `internal/jobs` also fails when run alone.

### DIFF
--- a/cmd/memory.go
+++ b/cmd/memory.go
@@ -71,17 +71,35 @@ func openMemoryStore() (*memorydb.Store, error) {
 	return store, nil
 }
 
-// wireImageRecorder opens the memory store and attaches it to the registry as an image recorder.
+type imageRecordStore interface {
+	tools.ImageRecorder
+	Close() error
+}
+
+type lazyImageRecorder struct {
+	open func() (imageRecordStore, error)
+}
+
+func (r *lazyImageRecorder) RecordImage(ctx context.Context, record *memorydb.ImageRecord) error {
+	store, err := r.open()
+	if err != nil {
+		return err
+	}
+	defer store.Close()
+	return store.RecordImage(ctx, record)
+}
+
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing in favor of a consolidated local implementation that also fixes late recorder wiring and adds focused lifecycle coverage.

PR #924, opened 2026-07-16
feat: add ephemeral side-question overlays

sam-saffron-jarvis:
## What

Add Claude-style ephemeral side questions to the TUI and web UI.

`/side <question>` opens a one-turn overlay over the current conversation. The main response keeps running and remains visible; the user is never switched into a second transcript.

Each side question receives:

- a fresh snapshot of the latest **completed** main context;
- up to 20 earlier private side question/answer pairs;
- no partial main assistant turn;
- no tools, MCP, native search, approvals, delegation, or subagents.

Side questions are runtime-local and intentionally absent from the durable main transcript, resume state, FTS, exports, compaction input, title generation, and message counts.

## TUI

- `/side <question>` remains available while main is streaming
- `/side` opens or reopens a compact ephemeral clarification thread with a pinned one-line composer
- follow-ups send directly from the overlay; each is still a fresh one-turn request with updated completed-main context
- chronological `You` / `Side` transcript replaces one-at-a-time history browsing
- centered responsive overlay leaves the main transcript visible underneath
- a 120×36 terminal gets a 22-row response viewport; taller terminals grow to 40 response rows
- rich Markdown is rendered with term-llm's shared ANSI renderer, including headings, tables, lists, and code
- the header reports `main responding` / `main running` while both streams are active
- `Esc` cancels an active side answer without affecting main; when idle it clears a side draft before closing
- `PgUp`/`PgDn` scroll the private transcript
- `Ctrl+C` copies the latest answer and confirmed `Ctrl+X` clears history
- side usage appears separately in `/stats`
- streaming slash submission clears the composer and invalidates the streaming viewport before rendering the overlay

There is no `/main`, `/side close`, child conversation, or multi-runtime navigation model.

## Web

- adds a responsive 1200px / 86vh reading overlay owned by the active session page rather than another session route
- includes a pinned side-only composer and chronological clarification transcript
- renders side answers through the existing sanitized assistant-Markdown pipeline
- streams main and side responses concurrently
- supports cancel, history navigation, copy, clear, and reconnect to the same live runtime
- private runtime endpoints inherit provider/model/reasoning configuration server-side
- persisted main history is hydrated before a side request after runtime eviction or restart
- unknown session IDs are rejected

## Context and isolation

- captures provider-safe context from the active main runtime
- excludes the active incomplete user/assistant boundary and malformed tool protocol fragments
- places the side policy in the leading privileged message block
- caps successful private history at 20 exchanges
- uses an independent cancellation context and generation guards
- prevents cancel/restart overlap even when a provider ignores cancellation
- converts attempted tool calls into synthetic warnings without dispatching them
- retains billable side usage in runtime/session statistics without polluting model context

The old persistent-side schema and lifecycle are removed. Schema v39 cleans databases that briefly ran the retired v38 branch migration.

## Tests

Coverage includes:

- completed-main snapshot boundaries and ordered tool sanitization
- rolling history and one-turn/tool-less request construction
- main/side cancellation isolation and stubborn-provider teardown
- TUI overlay controls, generation isolation, reasoning inheritance, and streaming composer/cache clearing
- web SSE lifecycle, backpressure, runtime hydration, nonexistent sessions, and server-side model inheritance
- SQLite transcript isolation and old-schema cleanup
- restored jobs cancellation, MCP locking, image-recorder lifecycle, and shell-approval regressions from the pre-rework branch

## Verification

- `gofmt -w .`
- `go build ./...`
- focused TUI, web, side-question, migration, tools, MCP, jobs, and image-recorder tests
- race tests for side runtimes and restored concurrency paths
- broad tests under isolated HOME/XDG directories
- `git diff --check`

The unrelated existing `internal/jobs.TestProgramRunnerDoesNotLeakBackgroundChildren` process-cleanup test remains flaky in this container and also fails independently of these changes.


SamSaffron:
implemented separately

PR #925, opened 2026-07-16
fix: misfire_policy=skip jobs run after service downtime

sam-saffron-jarvis:
## What changed

- Reconcile overdue `misfire_policy=skip` schedules synchronously when the jobs manager starts, before scheduler and worker goroutines begin.
- Advance skipped cron jobs to their next future occurrence without creating a run.
- Disable expired skipped once jobs without creating a run.
- Keep `misfire_policy=run` jobs overdue so the scheduler queues one normal catch-up run.
- Guard each persisted schedule transition and commit it transactionally.
- Add a file-backed restart test covering overdue cron and once jobs under both policies.

## Why this is high-value

Jarvis's bundled memory-maintenance and system-upgrade jobs explicitly use `misfire_policy=skip`. Previously the field was persisted but never read, so restarting after downtime immediately launched stale work despite that policy. This could duplicate memory maintenance, trigger delayed package upgrades, waste provider calls, or produce stale side effects during recovery.

## Validation

- `go build ./...`
- `go test ./cmd -run TestNewJobsV2ManagerReconcilesMisfirePolicies -count=1`
- Related jobs scheduler tests pass.
- `go test ./...` was run; the changed and related packages pass. The repository's unrelated `internal/jobs.TestProgramRunnerDoesNotLeakBackgroundChildren` timing test failed under the full-suite load but passed when rerun alone.
- `git diff --check`

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -702,6 +702,10 @@ func newJobsV2ManagerWithNotifier(dbPath string, workers int, llmExec serveJobsE
 		_ = db.Close()
 		return nil, err
 	}
+	if err := mgr.reconcileMisfiredSchedules(time.Now().UTC()); err != nil {
+		_ = db.Close()
+		return nil, err
+	}
 
 	mgr.wg.Add(1)
 	go mgr.schedulerLoop()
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by a direct maintainer-tree implementation. The fix now reconciles skipped misfires synchronously before scheduler startup, advances cron and disables once schedules transactionally, compare-and-swaps observed cron schedule inputs to avoid overwriting concurrent edits, validates skip/run policy values, and has deterministic plus file-backed startup coverage.

PR #933, opened 2026-07-18
fix: Isolated skill streaming shifts 1,024 retained events on every token

sam-saffron-jarvis:
## What changed

- Replaced isolated skill run history's per-event slice shift with bounded circular storage.
- Materialized retained events in chronological order only for snapshots and SSE replay.
- Extended the bounded-history test through more than four complete wraps, checking snapshot and replay ordering.
- Added a 10,000-event append benchmark.

## Why this is high-value

Isolated web skills forward token-sized child text deltas into this history. After 1,024 events, the old implementation copied all 1,024 retained interface-bearing events while holding the run mutex for every subsequent delta. That made a long response perform quadratic copying on the same lock used by SSE subscriptions, snapshots, cancellation, and completion.

The circular buffer replaces each post-limit 1,024-event copy with one indexed overwrite, preserving sequence numbers, the retention limit, chronological replay, and subscriber overflow behavior. On this host, the 10,000-event benchmark improved from roughly 6.83 ms and 244,642 B/op to 0.42 ms and 146,370 B/op (about 16x faster).

## Validation

- `gofmt -w cmd/serve_skills.go cmd/serve_skills_test.go`
- `go build ./...`
- Targeted skill history, overflow, and completion tests pass.
- Targeted race tests pass.
- `XDG_CONFIG_HOME=$(mktemp -d) go test ./... -skip '^TestProgramRunnerDoesNotLeakBackgroundChildren$'` passes.
- Full `go test ./...` was also run. Its skill-discovery tests require an isolated `XDG_CONFIG_HOME`; after isolating that config, the only failure was the unrelated pre-existing `TestProgramRunnerDoesNotLeakBackgroundChildren`, which reproducibly fails alone on this host because the background PID remains observable beyond its one-second deadline.
- `go test ./cmd -run '^$' -bench '^BenchmarkServeSkillRunAppendEvents$' -benchmem -count=3`
- `git diff --check`

### DIFF
--- a/cmd/serve_skills.go
+++ b/cmd/serve_skills.go
@@ -54,6 +54,8 @@ type serveSkillRun struct {
 	Activation        *skills.Activation
 	cancel            context.CancelFunc
 	events            []serveSkillRunEvent
+	eventHead         int
+	eventCount        int
 	nextSequence      int
 	subscribers       map[int]chan serveSkillRunEvent
 	subscriberDropped map[int]bool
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by a direct maintainer-tree implementation. The retained skill-event history now uses an O(1) circular overwrite without a redundant count invariant; snapshots materialize chronological order, while replay iterates the ring directly and allocates only matching events. Multi-wrap ordering/payload coverage and a benchmark were added.

PR #936, opened 2026-07-19
fix: Side-question streaming rebuilds the full response string on every delta

sam-saffron-jarvis:
## What changed

- Store the live side-question response as an appendable byte slice instead of an immutable string.
- Append text deltas into that buffer and reset its length on an attempt discard.
- Materialize a string only when taking a side-question view or backup, while releasing the buffer when the response is cleared.
- Add a benchmark that streams 10,000 one-byte deltas through `startSideQuestion`.

## Why this is high-value

Side questions are streamed on the user-facing web path. Previously, every delta used string concatenation while holding the side-question mutex, allocating and copying the full response accumulated so far. Token-sized deltas therefore caused quadratic copying, extra GC pressure, and progressively longer lock holds while the user was waiting for the answer.

Appending to a byte slice grows amortized linearly and keeps string materialization off the per-delta path. In the 10,000-delta benchmark, the fix reduced time from about 9.67 ms/op to 0.72 ms/op, allocated bytes from 53.3 MB/op to 137 KB/op, and allocations from 10,044/op to 52/op.

## Validation

- `gofmt -w cmd/serve_side_question.go cmd/serve_side_question_test.go`
- `go build ./...`
- `XDG_CONFIG_HOME="$(mktemp -d)" go test ./...`
- `go test ./cmd -run 'TestServeSideQuestion|TestReplaceHistoryFailureRestoresSideQuestionState' -count=1`
- `go test ./cmd -run '^$' -bench '^BenchmarkStartSideQuestionTextDeltas$' -benchmem -benchtime=5x -count=1`
- `git diff --check`

The isolated XDG config prevents locally installed user skills from affecting two unrelated skill-visibility tests.

### DIFF
--- a/cmd/serve_side_question.go
+++ b/cmd/serve_side_question.go
@@ -31,7 +31,7 @@ type sideQuestionRuntime struct {
 	reasoningEffort string
 	reasoningMode   string
 	question        string
-	response        string
+	response        []byte
 	synthetic       bool
 	usage           llm.Usage
 	totalUsage      llm.Usage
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing as superseded by a direct maintainer-tree implementation. Live side-question deltas now append to a byte buffer, discarded attempts release their backing storage, replacement/restore paths detach buffers, and views/backups materialize immutable strings. Live discard/UTF-8 behavior and the 10k-delta benchmark are covered.

PR #937, opened 2026-07-19
fix: Guardian serializes parallel shell approvals across full provider round-trips

sam-saffron-jarvis:
## What changed

- Added a bounded pool of three Guardian reviewers, each backed by an independently constructed provider.
- Dispatches concurrent policy reviews to separate reviewers instead of holding one reviewer's mutex across every provider round-trip.
- Reuses idle reviewers in LIFO order so ordinary sequential reviews continue using a warm delta-session cache.
- Added an installed-callback concurrency test with three independent delayed providers; all three reviews must enter their provider concurrently and finish in roughly one provider delay.

## Why this is high-value

Auto-approved shell calls are on Jarvis's normal agent path. Distinct first-time commands miss deterministic and exact-command approvals, and parallel tool scheduling can issue several such reviews together. Previously all of those reviews shared one `guardian.Reviewer`, whose mutex is intentionally held for conversation-state safety across the entire streamed provider response. Three parallel approvals therefore incurred roughly three sequential provider round-trips and could make the run appear stuck.

The pool preserves per-reviewer conversation serialization and provider isolation while allowing up to three common parallel misses to overlap. This changes their approval latency from approximately N provider delays to one provider delay for N <= 3, without changing deterministic approval fast paths or Guardian decisions.

## Validation

- `go build ./...`
- `go test -race ./internal/guardian ./cmd -run 'TestInstallGuardianReviewerCallbacksRunsParallelReviewsConcurrently|TestReviewerManagedProviderUsesFullThenDeltaTurns' -count=1`
- `go test ./cmd -run TestInstallGuardianReviewerCallbacksRunsParallelReviewsConcurrently -count=10`
- `git diff --check`

### DIFF
--- a/cmd/guardian_wiring.go
+++ b/cmd/guardian_wiring.go
@@ -16,6 +16,8 @@ import (
 
 var newGuardianProviderByName = llm.NewProviderByName
 
+const guardianReviewerPoolSize = 3
+
 func preflightHeadlessApproval(cfg *config.Config, resolved resolvedApprovalMode, providerName, modelName string) error {
 	if resolved.Mode != tools.ModeAuto {
 		return nil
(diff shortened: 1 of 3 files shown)


SamSaffron:
Closing as superseded by a more lifecycle-safe direct implementation. Guardian now uses a bounded pool of three independent reviewers, creates only the primary eagerly, expands lazily under contention, degrades safely if optional expansion fails, preserves LIFO warm-session reuse, supports context-aware waiting, and explicitly cancels/cleans pools on replacement and runtime shutdown. Concurrency, cancellation, bounds, replacement, and cleanup are tested.

PR #939, opened 2026-07-20
fix: Exited MCP servers remain permanently marked ready

sam-saffron-jarvis:
## What changed

- Start a watcher on each successfully initialized MCP `ClientSession` using the SDK's `ClientSession.Wait` API.
- When the active session terminates unexpectedly, clear that exact client's running session and cached tools, mark the server failed with the terminal error, and emit a failed status update.
- Guard the transition by both client and session identity so an old watcher cannot overwrite an explicit disable or a restarted server.
- Add an integration test whose stdio MCP helper exits during a tool call, then verify the server becomes failed, its tools disappear, a failed update is emitted, and `Restart` restores a working tool.

## Why this is high-value

Long-running TUI, web, Telegram, and jobs runtimes reuse one MCP manager. Previously, a stdio MCP process that crashed after initialization remained permanently `ready`, so every subsequent call failed while the dead server's tools continued to be advertised. Detecting the SDK session's terminal state makes the failure visible immediately and restores the existing safe restart path without retrying an ambiguous tool call.

## Validation

- `gofmt -w internal/mcp/client.go internal/mcp/manager.go internal/mcp/manager_test.go`
- `go build ./...`
- `go test -race ./internal/mcp -run 'TestManager(ExitedServerBecomesFailedAndCanRestart|Enable_ReadyStdioServerSurvivesStartupTimeoutContext|CallToolConcurrentDisable)' -count=5`
- `git diff --check`

### DIFF
--- a/internal/mcp/client.go
+++ b/internal/mcp/client.go
@@ -185,6 +185,34 @@ func (t *headerTransport) RoundTrip(req *http.Request) (*http.Response, error) {
 	return t.base.RoundTrip(req)
 }
 
+func (c *Client) currentSession() *mcp.ClientSession {
+	c.mu.RLock()
+	defer c.mu.RUnlock()
+	return c.session
+}
+
+// clearTerminatedSession clears a session only if it is still the active one.
+// This prevents a watcher for an old session from disrupting a restarted client.
+func (c *Client) clearTerminatedSession(session *mcp.ClientSession) bool {
+	c.mu.Lock()
+	if c.session != session || !c.running {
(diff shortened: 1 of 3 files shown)


SamSaffron:
Closing as superseded by a direct implementation. Each ready MCP session is watched through ClientSession.Wait; terminal sessions are cleared and marked failed only when manager, client, and session identities still match. Cached tools disappear, restart recovers, and explicit disable is covered to prevent stale watchers from emitting a false failure.

PR #946, opened 2026-07-22
fix: Mining limit permanently starves older unmined sessions

sam-saffron-jarvis:
## What changed

- Stop truncating completed-session candidates before mining state and pending messages are checked.
- Apply `--limit` to sessions that actually reach an extraction attempt, so fully mined sessions no longer consume the allowance.
- Add a regression test proving an older 51st candidate remains available when `--limit 50` is configured.

## Why this is high-value

The bundled `mine-sessions` job runs every 30 minutes with `--limit 50`. Completed sessions are ordered newest-first, so the previous candidate-level limit repeatedly selected the same newest 50 sessions even after they were fully mined. Any pending session older than that window could therefore remain unmined forever while every scheduled job reported success. Applying the limit after pending messages are confirmed lets the job scan past completed work and recover older memory backlogs or failed extractions.

## Validation

- `gofmt -w cmd/memory_mine.go cmd/memory_mine_test.go`
- `go build ./...`
- `go test ./cmd -run 'TestCollectMineCandidatesDoesNotApplyLimitBeforeMiningStateCheck|TestCollectInsightCandidatesAppliesLimitAfterSkippingMinedSessions'`
- `TEST_HOME=$(mktemp -d) && HOME="$TEST_HOME" XDG_CONFIG_HOME="$TEST_HOME/.config" CODEX_HOME="$TEST_HOME/.codex" go test ./...`
- `git diff --check`

### DIFF
--- a/cmd/memory_mine.go
+++ b/cmd/memory_mine.go
@@ -283,10 +283,15 @@ func runMemoryMine(cmd *cobra.Command, args []string) error {
 	}
 
 	var totalCreated, totalUpdated, totalSkipped int
+	var mineAttempts int
 	var summaryStats memoryMineSummaryStats
 	fragmentCache := newMemoryAgentFragmentCache(memoryMineTaxonomyMaxTokens)
 
 	for i, candidate := range candidates {
+		if memoryMineLimit > 0 && mineAttempts >= memoryMineLimit {
+			break
+		}
+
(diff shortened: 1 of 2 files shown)


SamSaffron:
Superseded by a clean combined implementation on current main. The replacement will preserve attempt-based --limit semantics while adding batched fully-mined-session pruning.

PR #947, opened 2026-07-22
fix: Memory mining rescans every completed session every 30 minutes

sam-saffron-jarvis:
## What changed

- Added a batched SQLite session-store capability that returns the maximum persisted message sequence for up to 500 requested sessions per query.
- Added a single-query memory mining-state listing and use it to filter fully mined sessions before loading full session rows.
- Carried the already-loaded mining state into the mining loop, eliminating the second per-candidate state lookup.
- Added focused coverage for 5,000 fully mined sessions, asserting candidate discovery performs one batch-capability call per store and zero `Get`, `GetState`, `GetMessages`, or `GetMessagesFrom` point reads. Added SQLite coverage across the 500-session batch boundary.

## Why this is high-value

Jarvis runs memory mining every 30 minutes and retains an ever-growing completed-session history. Previously, even when every old session was current, each run loaded every full session, queried its mining state, and issued an empty message query. That produced roughly three SQLite point reads per historical session on every run, plus avoidable session/message allocations.

The common fully-mined path now consists of paged summary listing, one mining-state scan, and batched indexed maximum-sequence lookups. Fully mined sessions are rejected before full session or message loading, so SQLite round trips, decoding, allocations, filesystem I/O, and read-transaction contention no longer grow at several point queries per historical session. Pending sessions preserve the existing loading and mining behavior.

## Validation

- `gofmt -w cmd/memory_mine.go cmd/memory_mine_test.go internal/memory/store.go internal/memory/store_test.go internal/session/store.go internal/session/sqlite.go internal/session/sqlite_test.go`
- `go build ./...`
- `git diff --check`

### DIFF
--- a/cmd/memory_mine.go
+++ b/cmd/memory_mine.go
@@ -61,7 +61,12 @@ When finished, reply briefly with plain text like "done".`
 type memoryMineCandidate struct {
 	Summary session.SessionSummary
 	Session *session.Session
+	State   *memorydb.MiningState
 	Agent   string
+}
+
+type memoryMiningStateStore interface {
+	ListStates(ctx context.Context) ([]memorydb.MiningState, error)
 }
 
 type memoryAgentFragmentStore interface {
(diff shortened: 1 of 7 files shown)


SamSaffron:
Superseded by a clean combined implementation on current main. The replacement will retain the batched pruning design while fixing --limit semantics, stale state handling, and current-main conflicts.

PR #954, opened 2026-07-23
fix: Gemini accepts truncated SSE responses as successful completions

sam-saffron-jarvis:
## What changed

- Track whether Gemini's primary candidate emitted an accepted terminal `finishReason` (`STOP` or `MAX_TOKENS`).
- Return a `StreamIncompleteError` when the SSE connection reaches EOF before that terminal reason, preventing the provider from emitting `EventDone` for partial output.
- Add a provider-level regression test for a response that emits text and then disconnects, and tighten complete-stream coverage to require `EventDone`.

## Why this is high-value

A proxy reset, provider disconnect, or dropped HTTP/2 stream could previously turn partial Gemini output into a successful completion. That silently corrupts the persisted/displayed answer and can omit trailing text or tool calls. Returning the existing typed incomplete-stream error routes the failure through term-llm's established transient retry and recovery behavior instead.

## Validation

- `gofmt -w internal/llm/gemini_api.go internal/llm/gemini_test.go`
- `go build ./...`
- `go test ./internal/llm -run 'Test(GeminiStreamUsesMinimalRESTClient|GeminiStreamRejectsTruncatedResponse|ReadGeminiSSE|GeminiStreamRejectsEmptyCandidateResponse)'`
- `HOME=$(mktemp -d) GOPATH=/home/agent/go GOCACHE=/home/agent/.cache/go-build go test ./...`
- `git diff --check`

### DIFF
--- a/internal/llm/gemini_api.go
+++ b/internal/llm/gemini_api.go
@@ -196,9 +196,13 @@ func streamGeminiResponses(ctx context.Context, client *http.Client, baseURL, ap
 
 func readGeminiSSE(r io.Reader, handle func(*geminiGenerateContentResponse) error) error {
 	decoder := newSSEDecoder(r, sseDecoderOptions{Transport: "Gemini SSE"})
+	sawTerminal := false
 	for {
 		_, data, err := decoder.Next()
 		if err == io.EOF {
+			if !sawTerminal {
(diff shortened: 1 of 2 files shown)


SamSaffron:
Integrated into the main working tree with centralized successful terminal-reason classification and clearer MAX_TOKENS test naming. Closing this superseded PR; provider regressions and the full suite pass.

PR #955, opened 2026-07-23
fix: Anthropic treats truncated streams as successful completions

sam-saffron-jarvis:
## What changed

- Track Anthropic's required `message_stop` terminal event in both standard and beta web-search streams.
- Return `StreamIncompleteError` when a syntactically valid SSE stream reaches EOF without that marker, instead of emitting `EventDone`.
- Delay beta server-tool fallback success until after terminal validation so an interrupted active tool is not reported as successful.
- Add standard and beta HTTP stream regressions covering partial output followed by clean EOF.

## Why this is high-value

Anthropic is a regular provider path, and its SDK treats clean EOF as a successful scanner termination even when the response was cut off. A proxy reset or provider disconnect could therefore persist partial text or tool activity as a completed answer. Classifying the missing terminal event as `StreamIncompleteError` activates the engine's existing retry/recovery path rather than silently corrupting the transcript.

## Validation

- `go test ./internal/llm -run 'TestAnthropicStreamsRejectMissingMessageStop|TestStreamWithSearchServerToolExecEnd|TestAnthropicStreamSerializesReasoningEffort' -count=1`
- `go build ./...`
- `XDG_CONFIG_HOME=<temporary-empty-directory> go test ./...`
- `git diff --check`

### DIFF
--- a/internal/llm/anthropic.go
+++ b/internal/llm/anthropic.go
@@ -281,6 +281,7 @@ func (p *AnthropicProvider) streamStandard(ctx context.Context, req Request) (St
 		}
 
 		var lastUsage *Usage
+		sawMessageStop := false
 		var streamOpts []option.RequestOption
 		if p.use1m {
 			streamOpts = append(streamOpts, option.WithHeaderAdd("anthropic-beta", the1mBetaHeader))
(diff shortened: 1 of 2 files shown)


SamSaffron:
Integrated into the main working tree and hardened for the shared Bedrock path: EOF is normalized only after message_stop validation, the broad EOF-to-Done wrapper was removed, and interrupted server tools now receive a failed end event. Closing this superseded PR.

PR #959, opened 2026-07-24
fix: Shell tracking rereads every previously touched file despite bounded affected_paths

sam-saffron-jarvis:
## What changed

- Treat non-empty `affected_paths` as the authoritative shell-tracking scope.
- Skip the `SessionPaths` lookup and historical-path append when bounded hints are present.
- Preserve the existing session-path and Git-status fallbacks for shell calls without hints.
- Add a regression test with 225 historical session files that verifies only the hinted file is snapshotted and recorded, and that `SessionPaths` is never queried.

## Why this is high-value

Shell tracking runs synchronously before and after every shell command. Previously, even a one-file `affected_paths` hint caused every file accumulated over a long-lived session to be statted and potentially read again. The per-call caps limited the damage but still allowed up to 200 content reads and 64 MiB of snapshot traffic, plus a second pass over historical candidates after command execution.

Trusting the already-documented bounded hints changes this work from O(session history) to O(declared scope), eliminating repeated database lookup, file I/O, and content-buffer allocation on Sam/Jarvis's frequent shell-tool path while retaining best-effort broad tracking when hints are absent.

## Validation

- `go build ./...`
- `go test ./internal/tools -run 'TestShellTool(AffectedPathsSkipSessionTrackedPaths|RecordsSessionTrackedPaths|RecordsAffectedPaths)$' -count=1`
- `go test ./internal/tools -run TestShellToolAffectedPathsSkipSessionTrackedPaths -count=10`
- Full `go test ./...` passed with an isolated `HOME`, `XDG_CONFIG_HOME`, and `CODEX_HOME` so host-global skills could not displace test fixtures.
- `git diff --check`

### DIFF
--- a/internal/tools/file_track_test.go
+++ b/internal/tools/file_track_test.go
@@ -18,11 +18,12 @@ import (
 
 // fakeFileRecorder captures ChangeRecords for assertions.
 type fakeFileRecorder struct {
-	mu           sync.Mutex
-	records      []filetrack.ChangeRecord
-	sessionPaths []string
-	seq          int64
-	maxFileBytes int // 0 = filetrack.DefaultMaxFileBytes
+	mu                sync.Mutex
+	records           []filetrack.ChangeRecord
+	sessionPaths      []string
+	sessionPathsCalls int
+	seq               int64
+	maxFileBytes      int // 0 = filetrack.DefaultMaxFileBytes
 }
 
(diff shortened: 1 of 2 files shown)


SamSaffron:
Integrated into the main working tree and hardened: affected_paths is now explicitly authoritative, blank hints are normalized so fallback tracking remains active, and the tool contract/comments document the bounded scope. Closing this superseded PR.

PR #960, opened 2026-07-24
fix: Transcript scans monopolize the sole session database connection

sam-saffron-jarvis:
## What changed

- Keep the existing single pooled writer connection for SQLite session mutations.
- Open one separately pooled, read-only connection for normal file-backed WAL stores.
- Route `GetTranscriptSnapshot` and `GetMessagesByTranscriptRanges` through that reader while preserving the single-connection fallback for `:memory:` and read-only stores.
- Close both handles with the store.
- Add an integration test that holds a transcript read transaction open and verifies `AddMessage` still completes through the writer within a one-second deadline.

## Why this is high-value

Transcript snapshot and range reads scan and decode complete message-part JSON while their read transaction remains open. Large, tool-heavy sessions therefore occupied the store's only pooled connection and forced unrelated persistence callbacks to wait before they could even begin a write transaction. In ordinary web, Telegram, and scheduled-job use, that head-of-line blocking can consume bounded persistence deadlines and delay or lose best-effort streaming checkpoints.

File-backed databases already use WAL, so one read-only connection can scan a coherent snapshot concurrently with the serialized writer. This removes connection-pool wait time for writes without adding concurrent writers, and caps the reader at one connection to bound SQLite cache and connection overhead. Raw `:memory:` stores retain their required single connection.

## Validation

- `go build ./...`
- Targeted SQLite connection and overlap tests pass.
- `go test ./...` passes with an isolated HOME/XDG config (preventing locally installed Jarvis skills from contaminating cmd skill-discovery fixtures).
- `git diff --check`

### DIFF
--- a/internal/session/sqlite.go
+++ b/internal/session/sqlite.go
@@ -22,6 +22,7 @@ import (
 // SQLiteStore implements Store using SQLite.
 type SQLiteStore struct {
 	db                       *sql.DB
+	readDB                   *sql.DB
 	cfg                      Config
 	hasGeneratedTitles       bool // true if sessions table has generated title columns
 	hasCompactionSeq         bool // true if sessions table has compaction_seq column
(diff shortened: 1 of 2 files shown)


SamSaffron:
Integrated into the main working tree and hardened: transcript readers now use URI-safe paths, are created only when WAL is actually active, and read-only DSNs no longer attempt journal-mode mutation. Added special-character path coverage. Closing this superseded PR.

PR #970, opened 2026-07-27
feat(hub): add example systemd unit under ops/

arpitjalan:
Adds an example systemd unit at `ops/systemd/term-llm-hub.service` (plus a short docs section) for running the Hub as a background service, following the `ops/nginx/` precedent — no CLI machinery, no Go changes. The unit is sandboxed by default (`DynamicUser=`, no home access), keeps tokens in `/etc/term-llm-hub.env` (0600) via `EnvironmentFile=` so they never appear in the unit or process listings, and includes a commented `User=root` variant for auto-discovering local `contain` workspaces (with a note on why `User=` is required: systemd only sets `$HOME` when it's present). Replaces the earlier `serve hub systemd` generator approach per review feedback.


arpitjalan:
@SamSaffron this PR is meant for my term-llm instance that is running on a DO droplet. Does the direction looks correct to you?

SamSaffron:
not sure about this, cant we just have the unit do `term-llm serve hub` the systemd is just an example way of how to install ... can be an example directory or something ... building the machinery into term llm does not feel right to me.  

also droplet wise wouldn't a package be a better direction, running as root also makes me very nervous

arpitjalan:
Yeah fair, the machinery was overkill - reworked it as a commented example at `ops/systemd/term-llm-hub.service` (same idea as the nginx configs), the unit just runs `term-llm serve hub`.

On root: the example is sandboxed by default now (`DynamicUser=`, no home access), root is only a commented-out variant for people who want local `contain` workspaces auto-discovered (it needs their 0600 `.env` files + loopback).

Agree a package is the right direction for servers longer-term. 👍

SamSaffron:
sure, this is good

PR #981, opened 2026-07-29
fix: Single tool calls can make cancellation and shutdown hang forever

sam-saffron-jarvis:
## What changed

- Run single and serial tool executions behind a buffered result channel, allowing the engine to synthesize a paired cancellation result without waiting for a context-ignoring tool.
- Apply the same buffered goroutine/select pattern to synchronous CLI-bridge tool execution while preserving panic recovery and normal tool output handling.
- Add regression coverage for a fake tool that starts and then blocks independently of its context, including direct single-call cancellation, `Engine` stream shutdown, and synchronous bridge execution.

## Why this is high-value

Tool execution is central to Jarvis's web, Telegram, and CLI workflows. A wedged custom tool, MCP transport, subprocess, or network client previously held the engine goroutine forever even after cancellation. Because `channelStream.Close` waits for that goroutine, one non-cooperative tool could permanently strand a response run and block runtime cleanup or service shutdown. The new cancellation boundary lets the engine and stream owner finish promptly while a late tool return is safely absorbed by the buffered channel.

## Validation

- `gofmt -w internal/llm/engine.go internal/llm/engine_test.go`
- `go build ./...`
- `XDG_CONFIG_HOME=$(mktemp -d) go test ./... -count=1`
- `go test -race ./internal/llm -run 'TestExecuteToolCallsSingleReturnsOnContextCancel|TestEngineStreamCloseReturnsWhenSingleToolIgnoresCancellation|TestHandleSyncToolExecutionReturnsOnContextCancel' -count=10`
- `git diff --check`

### DIFF
--- a/internal/llm/engine.go
+++ b/internal/llm/engine.go
@@ -3567,9 +3567,10 @@ func (e *Engine) executeToolCalls(ctx context.Context, calls []ToolCall, paralle
 		return cancelledToolCallMessages(calls, err), nil
 	}
 
-	// Fast path: single call, no concurrency overhead
+	// Single calls still run asynchronously so cancellation cannot be held
+	// hostage by a non-cooperative tool implementation.
 	if len(calls) == 1 {
-		return e.executeSingleToolCallSafe(ContextWithApprovalTranscript(ctx, transcript), calls[0], send, debug, debugRaw)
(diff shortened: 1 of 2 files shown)


sam-saffron-jarvis:
Closing as a duplicate of #980. The nightly scanner emitted the same root cause twice under different wording; the dedupe pipeline has been hardened to use source-site identity as well as semantic overlap.

PR #991, opened 2026-08-03
Add /undo and /redo to web and TUI

sam-saffron-jarvis:
## What

Add `/undo` and `/redo` to both the TUI and web chat interfaces.

- `/undo` removes the latest real user message and the complete transcript suffix after it
- `/redo` restores that suffix exactly, including stable message IDs, structured parts, response identity, and timestamps
- no confirmation dialog
- restores the undone user text to the composer
- stores a durable per-session redo stack in SQLite so repeated `/undo` and `/redo` work across process restarts and clients
- rejects stale-client mutations with transcript revision and head-row checks
- refuses mutations while a response, compaction, or side question is active
- preserves compaction boundaries and excludes synthetic compaction rows
- resets provider continuation/runtime state after transcript mutation
- adds a reusable accessible web toast system for success and error feedback

Undo changes conversation history only. It does not reverse tool calls, file edits, shell commands, or other external side effects. Attachments remain available through `/redo` but are not reconstructed in the composer; the UI warns when this applies.

## Why

A mistaken prompt should be immediately editable without branching or manually rebuilding conversation context. Redo is included because an unconfirmed undo without recovery is a foot-gun.

## Storage and consistency

- Adds schema migration 44 with a durable per-session LIFO redo stack
- Supports `/undo /undo /redo /redo` in exact reverse order; ordinary transcript writes clear the whole redo stack
- Performs undo/redo in serialized SQLite transactions
- Bumps `transcript_rev` exactly once per successful mutation
- Invalidates redo on ordinary transcript mutation
- Preserves lifetime usage counters while reconciling transcript-derived metadata and the authoritative plan projection
- Clears durable provider continuation state to prevent an undone turn from leaking back through provider-side response chaining

## Verification

Passed with an isolated temporary home:

```sh
go build ./...
go test ./...
go vet ./...
```

Also passed focused race coverage for concurrent undo and optimistic mutation checks.

The implementation received an Opus merge-gate review. Its five P1 findings and eight P2 findings were fixed with regression coverage before the final full test run.

### DIFF
--- a/cmd/serve_handlers.go
+++ b/cmd/serve_handlers.go
@@ -1605,6 +1605,20 @@ func (s *serveServer) handleSessionByID(w http.ResponseWriter, r *http.Request) 
 		return
 	}
 
+	if suffix == "runtime/undo" || suffix == "runtime/redo" {
+		if r.Method != http.MethodPost {
+			w.Header().Set("Allow", "POST")
+			writeOpenAIError(w, http.StatusMethodNotAllowed, "invalid_request_error", "method not allowed")
+			return
+		}
+		if err := requireJSONContentType(r); err != nil {
+			writeOpenAIError(w, http.StatusUnsupportedMediaType, "invalid_request_error", err.Error())
+			return
+		}
+		s.handleSessionUndoRedo(w, r, sessionID, suffix == "runtime/redo")
+		return
+	}
+
 	if suffix == "runtime/effort" {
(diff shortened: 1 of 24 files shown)


SamSaffron:
Integrated locally into main as merge commit 5a5b2586 after resolving the app-send.js conflict against the trimmed dependency imports. Full build, test, and vet checks pass.

PR #992, opened 2026-08-03
fix: Telegram callback write failures bypass transcript reconciliation

sam-saffron-jarvis:
## What changed

- Return callback persistence errors from `runStoreOpWithoutCancel` while preserving its existing logging and timeout behavior.
- Mark the Telegram callback store queue degraded when a queued operation fails, so the existing final `ReplaceMessages(callback_reconcile)` path repairs the complete in-memory transcript.
- Add a queue unit test for operation errors and an integration test that fails the first assistant callback write, then verifies one reconciliation produces the exact user/assistant/tool/assistant transcript without duplicates.

## Why this is high-value

Telegram disables runtime persistence and relies on callback writes for assistant and tool messages. Previously, a transient callback `AddMessage` failure was only logged: the message was already present in the in-memory `produced` transcript, so the fallback did not run, while the queue remained healthy and skipped final reconciliation. That could permanently leave Telegram's durable session behind the answer Sam already received, causing stale context after restart or carryover. Treating any queued write error as degradation activates the already-established transcript repair path with a small behavioral change.

## Validation

- `gofmt -w internal/serve/telegram.go internal/serve/telegram_queue_test.go internal/serve/telegram_test.go`
- `go build ./...`
- `go test -race ./internal/serve -run 'TestTelegramStoreOpQueueOperationErrorDegrades|TestStreamReply_ReconcilesFailedCallbackPersistence' -count=1`
- `git diff --check`
- Ran `go test ./...`; all affected/relevant packages passed, but the repository-wide run still reports unrelated environment/baseline failures in two `cmd` skill-discovery tests (the host skill registry shadows test fixtures) and `internal/serveui`'s existing JS line-budget check (`20504` vs `20467`).

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -1103,9 +1103,9 @@ func (m *telegramSessionMgr) runStoreOpWithTimeout(sessionID, op string, fn func
 	}
 }
 
-func (m *telegramSessionMgr) runStoreOpWithoutCancel(ctx context.Context, sessionID, op string, fn func(context.Context) error) {
+func (m *telegramSessionMgr) runStoreOpWithoutCancel(ctx context.Context, sessionID, op string, fn func(context.Context) error) error {
 	if m.store == nil || fn == nil {
-		return
+		return nil
 	}
 	if ctx == nil {
 		ctx = context.Background()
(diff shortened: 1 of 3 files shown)


SamSaffron:
Closing because #993 supersedes this narrower machine PR, and the callback-write repair is included in the reviewed, improved local working-tree implementation. The combined changes are intentionally still uncommitted.

PR #993, opened 2026-08-03
fix: Telegram turn-write failures bypass final transcript reconciliation

sam-saffron-jarvis:
## What changed

- Made Telegram store helpers report whether writes succeeded.
- Marked a turn persistence-degraded when direct system/user/fallback transcript writes fail.
- Marked the callback persistence queue degraded when an enqueued operation executes but returns an error.
- Reused the existing final `ReplaceMessages` path whenever either direct or queued persistence degraded, including recovery of a failed first system-prompt write.
- Added focused coverage for direct user-write failure, queued assistant-write failure, complete exactly-once system/user/assistant/tool reconciliation, and queue degradation after an operation error.

## Why this is high-value

Telegram intentionally disables the shared runner's runtime persistence and relies on these per-turn writes. Previously, a transient direct or callback write failure was only logged, so a successful response could be delivered while durable history silently omitted the request, answer, system prompt, or tool context. That stale transcript would then be reused after restart or session carryover. The in-memory turn already contains the complete transcript, so final replacement is a narrow, safe repair.

## Validation

- `go build ./...`
- `go test ./internal/serve -run 'TestTelegramStoreOpQueue|TestStreamReply_ReconcilesTranscriptAfterTurnWriteFailure' -count=1`
- `go test -race ./internal/serve -run 'TestTelegramStoreOpQueue|TestStreamReply_ReconcilesTranscriptAfterTurnWriteFailure' -count=1`
- `go test ./...` (all Go packages pass except the unrelated existing `internal/serveui` source-size ratchet: 20,504 legacy JS lines versus its 20,467 threshold; this PR does not touch JS)
- `git diff --check`

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -1080,16 +1080,18 @@ func closeTelegramSessionWithTimeout(sess *telegramSession, wait time.Duration) 
 	cleanup()
 }
 
-func (m *telegramSessionMgr) runStoreOp(ctx context.Context, sessionID, op string, fn func(context.Context) error) {
+func (m *telegramSessionMgr) runStoreOp(ctx context.Context, sessionID, op string, fn func(context.Context) error) bool {
 	if m.store == nil || fn == nil {
-		return
+		return false
 	}
 	if ctx == nil {
 		ctx = context.Background()
 	}
 	if err := fn(ctx); err != nil {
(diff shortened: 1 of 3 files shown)


SamSaffron:
Closing this machine PR to consolidate its fix into a reviewed, improved local working-tree implementation. This also includes the overlapping behavior from #992; the combined changes are intentionally still uncommitted.

PR #996, opened 2026-08-04
feat: add Claude-style @ file autocomplete

sam-saffron-jarvis:
## What

Adds project-scoped `@` file and directory autocomplete to the chat TUI, following Claude Code 2.1.220's submit-time mention semantics.

Typing `@` fuzzy-finds project paths. Selection inserts ordinary text:

```text
@internal/mentions/parse.go
@"docs/file with spaces.md"
@internal/mentions/
```

The original user text remains intact. At submission, both selected and manually typed valid mentions are parsed and may produce bounded attachment context.

## Claude-compatible attachment behavior

The implementation was checked against the embedded JavaScript extracted from the official `@anthropic-ai/claude-code-linux-x64@2.1.220` npm binary.

- Non-PDF text files whose **total file size is at most 262,144 bytes** are eligible for eager attachment.
- Larger files attach nothing; the textual `@path` remains for the model to inspect with tools.
- A requested `#Lx` or `#Lx-Ly` range does not bypass the total-file-size gate.
- Attached text has a **25,000 locally estimated token** ceiling.
- Token/range overflow retries from the requested offset with at most **2,000 lines**.
- If fallback still exceeds the ceiling, attachment is omitted without blocking the prompt.
- Directories attach a non-recursive names-only listing capped at **1,000 entries**, followed by the exact `… and N more entries` marker when truncated.
- Quoted mentions are extracted before unquoted mentions and duplicate handling follows Claude's raw mention strings.
- Mention start boundaries include whitespace plus `。`, `、`, `？`, and `！`, matching 2.1.220.
- Denied, missing, binary, oversized, unsupported image/PDF, and unsafe paths remain textual references and do not block sending.

Intentional divergences:

- term-llm uses local token estimation rather than Claude's conditional token-count API call.
- Local image mentions and PDF inline/reference handling remain on term-llm's existing explicit attachment paths.
- term-llm confines eager reads to the active project and rejects UNC/network/root escapes rather than adding Claude's trusted-network-directory feature.
- Claude's already-read cache optimization, MCP resource templates, and `inputMentionsOnly` orchestration do not have relevant term-llm equivalents in this V1.

## Discovery and search

- Warm asynchronous project index; no filesystem traversal per keystroke.
- Whole-relative-path fuzzy matching with bounded top-K results.
- Git repositories use tracked plus non-ignored untracked files via `git ls-files`.
- Git discovery fails closed when ignore evaluation fails.
- Plain directories include dotfiles, exclude VCS metadata, and do not follow symlinks.
- Queries are cancellable and guarded by root, index generation, request ID, token text/range, and cursor position.
- Popup works in inline and alt-screen modes with fuzzy highlighting, wrapping navigation, Enter/Tab selection, and Escape dismissal.

## Read safety

Mention reads do not grant permissions or invoke interactive approval UI.

The actual opened descriptor is confined with Go's descriptor-relative `os.Root` implementation. The code securely stats, approves, opens, verifies with `os.SameFile`, and reads/lists from the validated descriptor. Regression tests cover final-component and parent-component symlink replacement after approval.

Unsupported platforms without a secure primitive fail closed.

## Compatibility

- Existing `/file` and image attachment behavior is unchanged.
- Original user text precedes generated mention context, including streaming interjections.
- Oversized or failed mentions never prevent submission.
- Enter falls through to ordinary submission when there is no current valid autocomplete selection.
- No new persistence schema or structured mention identity is introduced.

## Performance

Local i9-14900K matcher benchmarks over 100,000 candidates:

- selective query: ~0.8 ms/op
- one-character query: ~3.0 ms/op
- broad query: ~3.8 ms/op

## Verification

```sh
git diff --check
go test -race ./internal/mentions ./internal/tools ./internal/tui/chat
go test ./...
go build ./...
go vet ./...
```

All pass with isolated HOME/XDG directories.

Additional coverage includes:

- 262,144-byte boundary and 10 MB omission
- giant one-line bounded reads
- ranged mentions on oversized files
- 25,000-token fallback to 2,000 lines
- quoted paths, manual mentions, raw duplicate semantics and CJK boundaries
- 1,000-entry directory cap
- denied/root/symlink/UNC escape paths
- stale autocomplete state and cursor/query changes
- `/file`, images, interjections, history, FTS and export compatibility

Set `TERM_LLM_AT_MENTIONS=0` to disable the feature.

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,6 +61,19 @@ term-llm ask --provider zen "explain git rebase"
 term-llm chat
 ```
 
+### Project `@` mentions
+
+In chat, type `@` to fuzzy-find files and directories under the active project/worktree; the picker keeps Git ignore behavior, and Enter or Tab inserts ordinary text (`@path`, `@"path with spaces"`, or `@directory/`). Manually typed valid mentions work identically. Set `TERM_LLM_AT_MENTIONS=0` to disable the picker.
+
+Mentions are parsed only when the prompt is submitted, at the start of text or after whitespace, `。`, `、`, `？`, or `！`; these punctuation characters start mentions but do not terminate an unquoted path. The visible user text is not rewritten, duplicate references remain visible, and identical raw mention payloads attach once. Textually different references to the same path (for example, `@a` and `@./a`) remain distinct, matching Claude Code, and quoted matches attach before unquoted matches. `#L10` and `#L10-20` select lines. Resolution is confined to the active project/CWD, follows existing non-interactive read policy, never opens approval UI, and never grants read or write access. Missing, denied, escaped, binary, oversized, or otherwise failed resources remain textual references and do not block sending.
+
+Submit-time limits closely follow Claude Code 2.1.220:
+
+- Non-PDF text files are eligible only when their **total** size is at most 256 KiB; a line range does not bypass this gate.
+- Attached content is capped at 25,000 estimated tokens. On overflow, term-llm retries 2,000 lines from line 1 or the requested starting line; if that still exceeds the ceiling, nothing is attached. Unlike Claude Code's API-assisted exact count, term-llm deliberately uses its local four-bytes-per-token estimate, avoiding an API round trip.
+- Directories attach a live, non-recursive, names-only listing with at most 1,000 names and an `… and N more entries` marker.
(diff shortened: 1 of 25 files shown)


SamSaffron:
implemented in working tree and pushed

PR #998, opened 2026-08-04
feat: add authorized @agent delegation mentions

sam-saffron-jarvis:
## What

Adds explicit `@agent:<name>` delegation mentions to the chat TUI on top of the existing project file mention picker.

Examples:

```text
Ask @agent:codebase to trace the parser
Have @agent:reviewer challenge the proposed fix
Review @internal/mentions/parse.go with @agent:reviewer
```

Selecting an agent inserts ordinary editable text. It does **not** launch a child directly. At submission, term-llm revalidates the mention against the current runtime and adds provider-only delegation metadata asking the active model to call `spawn_agent` with a focused prompt.

Bare `@name` remains file/text syntax. `/handover` and startup `term-llm chat @agent` retain their existing active-agent-switch semantics.

## Runtime authorization

Agent completion is intentionally stricter than Claude Code's implementation. A candidate appears only when all of these are true:

- `spawn_agent` is registered in the current ToolManager;
- the current live Engine allowed-tools filter permits it;
- the actual `SpawnAgentTool` has a runner;
- current depth is below its maximum;
- its exact `allowed_agents` whitelist permits the name, with the existing empty-means-unrestricted semantics;
- the runner catalog resolves and validates the agent definition.

`SpawnAgentTool.Execute`, completion, and submit validation share the same local policy helper. The actual model-emitted tool call is still rechecked by both Engine and SpawnAgentTool.

Capability checks follow live model/effort Engine replacement and skill tool-filter changes rather than capturing the construction-time Engine.

## Syntax and UI

- `@agent:<query>` shows agents only.
- Path-like queries show files only.
- General `@<query>` shows categorized permitted agents followed by files/directories.
- Agent and path rankings remain independent.
- Category headers are not selectable.
- Stale rows are guarded by root, token, cursor, generation, request, and live capability checks.
- Trailing prose punctuation terminates an agent name.
- Quoted names are supported where the registry lookup grammar permits them.
- Unknown valid explicit names fail closed with an actionable footer error while preserving the draft and attachments.
- Agent-like text inside collapsed paste payloads cannot create hidden delegation intent.

## Provider and persistence semantics

The original visible request remains unchanged. Valid mentions create structural `PartAgentMention` metadata that becomes a provider-only delegation instruction at the provider conversion boundary.

It is excluded from:

- user bubbles and sticky previews;
- `TextContent` and FTS/history search;
- prompt history and restored drafts;
- Markdown and HTML exports;
- resume rendering.

Normal sends and streaming interjections use the same validation and ordering. The TUI never calls the runner or `SpawnAgentTool.Execute` directly, preserving ordinary tool transcript entries, progress events, cancellation, timeout/semaphore handling, parent/child session linkage, and result handling.

## Claude Code comparison

Claude Code 2.1.220 also treats configured-agent mentions as model-mediated intent rather than direct execution. It inserts `@"<type> (agent)"` and adds a reminder asking Claude to invoke the Agent tool.

This implementation uses the clearer `@agent:<name>` namespace and improves eligibility: Claude autocomplete checks only active-definition existence and can suggest agents the runtime later rejects; term-llm offers only agents the current runtime can actually spawn.

## Opus review

A post-implementation `claude-bin:opus-max` review found one blocker, two high, five medium, and six low issues.

Fixed findings include:

- hidden delegation instructions rendering inside the user bubble;
- capability pinned to an obsolete Engine after model/effort changes;
- punctuation and collapsed-paste false positives;
- quoted completion/submission asymmetry;
- skill side effects occurring before validation;
- dropped footer-clear commands and auto-send stalls;
- duplicate catalog validation, selection retention across category changes, empty-name policy bypass, and concurrent registry reads.

The only documented deferrals are broader product asymmetries: agent registry discovery remains rooted like the existing spawn runner after worktree switches, and `@agent` is currently a chat-TUI surface rather than Web/ask syntax.

## Verification

Passed with isolated HOME/XDG state:

```text
gofmt
git diff --check
go test -race ./cmd ./internal/agents ./internal/llm ./internal/mentions ./internal/render/chat ./internal/session ./internal/tools ./internal/tui/chat
go test ./...
go build ./...
go vet ./...
```

Focused deterministic regressions were repeated with `-count=10` and `-count=20`, including live Engine allow→deny and deny→allow transitions, skill filters, whitelist/depth/catalog policy, parser/paste cases, renderer/export/session round trips, interjection ordering, and categorized popup stale-state handling.

### DIFF
--- a/cmd/agents.go
+++ b/cmd/agents.go
@@ -293,9 +293,9 @@ func runAgentsList(cmd *cobra.Command, args []string) error {
 func runAgentsNew(cmd *cobra.Command, args []string) error {
 	name := args[0]
 
-	// Validate name
-	if strings.ContainsAny(name, "/\\:*?\"<>|") {
-		return fmt.Errorf("invalid agent name: %s", name)
+	// Validate the exact registry lookup key used by discovery and delegation.
+	if !agents.IsLookupName(name) {
+		return fmt.Errorf("invalid agent name %q: use at most %d letters, digits, underscores, or non-trailing spaces, dots, and hyphens", name, agents.MaxLookupNameRunes)
 	}
 
 	// Determine base directory
@@ -494,9 +494,9 @@ func runAgentsCopy(cmd *cobra.Command, args []string) error {
 	srcName := args[0]
 	destName := args[1]
 
-	// Validate dest name
-	if strings.ContainsAny(destName, "/\\:*?\"<>|") {
-		return fmt.Errorf("invalid agent name: %s", destName)
+	// Validate the exact registry lookup key used by discovery and delegation.
+	if !agents.IsLookupName(destName) {
+		return fmt.Errorf("invalid agent name %q: use at most %d letters, digits, underscores, or non-trailing spaces, dots, and hyphens", destName, agents.MaxLookupNameRunes)
 	}
 
 	cfg, err := loadConfigWithSetup()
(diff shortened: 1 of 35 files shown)


SamSaffron:
Brought these changes into the current working tree as uncommitted changes and verified them locally. Closing as requested.

PR #1002, opened 2026-08-05
fix: Stopping the agy proxy can nil its transport under an active request

sam-saffron-jarvis:
## What changed

- Snapshot the agy proxy's transport under `Server.mu` when `forward` begins, then use that stable pointer for the upstream round trip.
- Return a service-unavailable response if a handler starts after shutdown has already cleared the transport.
- Add a coordinated regression test that blocks generation request-body processing, calls `Stop`, releases the body, and verifies the active forward completes without a panic.

## Why this is high-value

`Stop` previously cleared `s.transport` while an already-running generation handler could still be reading, expanding, filtering, or tracing its request body. The handler then dereferenced the now-nil transport, producing both a data race and a possible nil-pointer panic. Because agy isolation is stopped during per-turn cleanup, ending or cancelling an agy-backed turn could crash the entire long-running Jarvis process rather than only failing that provider call.

The fix is deliberately lifecycle-local: active handlers retain the transport they started with, while handlers that lose the race with shutdown fail safely.

## Validation

- `go test -race ./internal/agyproxy`
- `go build ./...`

The required unisolated `go test ./...` run also completed all packages, but two unrelated `cmd` skill-discovery tests failed because the host's 30 user-global skills exhausted their configured metadata limits before the temporary project skill was listed. Isolating `XDG_CONFIG_HOME` removed that host configuration and the full suite passed.

### DIFF
--- a/internal/agyproxy/proxy.go
+++ b/internal/agyproxy/proxy.go
@@ -277,6 +277,14 @@ func (a loopbackAddr) Network() string { return "tcp" }
 func (a loopbackAddr) String() string  { return string(a) }
 
 func (s *Server) forward(w http.ResponseWriter, req *http.Request) {
+	s.mu.Lock()
+	transport := s.transport
+	s.mu.Unlock()
+	if transport == nil {
+		http.Error(w, "agy tool-filter proxy is stopping", http.StatusServiceUnavailable)
+		return
+	}
+
 	req.RequestURI = ""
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing this machine PR to consolidate its fix into a reviewed, improved local working-tree implementation. The combined changes are intentionally still uncommitted.

PR #1003, opened 2026-08-05
fix: Shutdown can strand a newly claimed job without executing it

sam-saffron-jarvis:
## What changed

- Serialize run admission with manager shutdown under the existing manager mutex.
- If shutdown wins after a worker commits a claim but before execution is admitted, return that still-claimed run to `queued` and clear its `worker_id` instead of leaving it stranded.
- Expand the shutdown regression test to close and reopen a file-backed jobs database, verify the run remains queued, and prove it can be reclaimed and executed successfully.

## Why this is high-value

Jobs are part of Jarvis's daily path. The race window is narrow, but previously a reload or shutdown in that window persisted the run as `claimed`; startup recovery then reported a misleading worker-lost failure, and the default single-attempt policy permanently skipped the occurrence. The fix keeps an unstarted occurrence executable after restart without changing normal execution or cancellation behavior.

## Validation

- `go build ./...`
- `go test ./cmd -run '^TestJobsV2ExecuteRunRequeuesClaimWhenManagerClosed$' -count=1`
- `go test -race ./cmd -run '^TestJobsV2ExecuteRunRequeuesClaimWhenManagerClosed$' -count=10`
- `git diff --check`

A plain `go test ./...` initially encountered two unrelated skill-discovery assertions because this host's 30 user-level Jarvis skills displaced the temporary project skill from tests capped at 8/10 visible skills. Isolating `XDG_CONFIG_HOME`, as in a clean test environment, made the full suite pass.

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -1418,7 +1418,12 @@ func (m *jobsV2Manager) executeRun(run jobsV2Run) {
 	}
 	ctx, cancel := context.WithTimeout(context.Background(), timeout)
 	m.mu.Lock()
-	closed := m.closed
+	if m.closed {
+		m.mu.Unlock()
+		cancel()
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing this machine PR to consolidate its fix into a reviewed, improved local working-tree implementation. The combined changes are intentionally still uncommitted.

PR #1013, opened 2026-08-07
fix: A stopping widget can erase and orphan its concurrent replacement

sam-saffron-jarvis:
## What changed

- Serialize each widget's complete stop lifecycle with the same mutex used for starts, including process termination, waiting, and Unix-socket cleanup.
- Use a lock-aware stop helper from startup cancellation paths that already hold the lifecycle mutex.
- Make process-exit cleanup conditional on the exiting command still owning the entry's recorded process.
- Add a regression test with a socket widget that delays SIGTERM handling, overlaps a stop with a restart request, and verifies the replacement remains tracked, retains its socket/proxy, serves successfully, and is the only surviving process.

## Why this is high-value

Idle reaping and manual stops can overlap a new widget request during the old process's termination window. Previously the request could start a replacement, then the old process's waiter could clear the replacement's process and proxy while deferred socket cleanup unlinked its live socket. That left a user-visible widget broken and its replacement process orphaned until operator cleanup or a server restart. Serializing stop and start closes this realistic lifecycle race, while the ownership check prevents any stale waiter from mutating newer process state.

## Validation

- `go build ./...`
- `go test -race ./internal/widgets -count=1 -timeout=45s`
- `HOME=<temporary> XDG_CONFIG_HOME=<temporary> go test ./...` (isolated from the host's installed skill catalog so environment-sensitive skill visibility tests use only their fixtures)
- `git diff --check`

### DIFF
--- a/internal/widgets/manager.go
+++ b/internal/widgets/manager.go
@@ -304,7 +304,7 @@ func (m *Manager) ensureRunning(e *widgetEntry) error {
 		return err
 	}
 	if m.isShuttingDown() {
-		e.stopProcess()
+		e.stopProcessLocked()
 		return errWidgetManagerShuttingDown
 	}
 	return nil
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #1016, opened 2026-08-08
fix: Stateful web runs continue after session persistence fails

sam-saffron-jarvis:
## What changed

- Fail stateful runs before `onStart` or provider execution when a configured session store cannot create or hydrate the requested session.
- Preserve existing execution behavior for explicitly stateless runs and runtimes with persistence disabled.
- Add regression coverage for failing `Get`/`Create` calls, callback suppression, and provider suppression, plus companion coverage for allowed non-persistent runs.

## Why this is high-value

The web/Jarvis path relies on stateful session persistence for both transcript durability and prior-turn context. Previously, a transient SQLite lock or I/O failure could still start and complete a provider run while every persistence and hydration path remained disabled. The answer could appear successful but disappear after reload, and resumed conversations could reach the provider without their stored history. Failing before the run starts prevents a temporary database fault from becoming permanent loss of the entire turn or a context-free response.

## Validation

- `go test ./cmd -run '^TestServeRuntime(StatefulRunRequiresSessionPersistence|NonPersistentRunsStillExecute)$'`
- `go build ./...`
- `HOME="$TEST_HOME" XDG_CONFIG_HOME="$TEST_HOME/config" XDG_DATA_HOME="$TEST_HOME/data" XDG_CACHE_HOME="$TEST_HOME/cache" GOMODCACHE="$GOMODCACHE" go test ./...`
- `go vet ./...`
- `git diff --check`

### DIFF
--- a/cmd/serve_runtime.go
+++ b/cmd/serve_runtime.go
@@ -1023,6 +1023,7 @@ func serveRuntimeSetupFromContext(ctx context.Context) func(*llm.Request) error 
 var (
 	errServeSessionBusy         = errors.New("session is busy processing another request")
 	errServeSessionLimitReached = errors.New("session limit reached: all sessions are busy")
+	errServeSessionPersistence  = errors.New("failed to persist or hydrate session")
 )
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #1017, opened 2026-08-08
fix: File-history total-size cap is never enforced during a long-running process

sam-saffron-jarvis:
## What changed

- Enforce `MaxTotalBytes` during ordinary `RecordChange` traffic instead of only during startup GC.
- Amortize the lightweight size check: run it after at most 8 MiB of retained input (using a proportional interval for smaller caps) or 64 metadata-only records.
- Run pruning in the record transaction so an enforcement failure rolls back the new record rather than reporting a committed write as failed, and invalidate per-session byte caches after a live check.
- Add a regression test that grows file history across sessions without calling `GC`, then verifies the database returns under the configured cap while the newest session survives and the oldest is pruned.

## Why this is high-value

Jarvis keeps a process-wide file-history store open across many sessions, while the previous total-budget enforcement ran only when that store was opened. A long-running serve/Jarvis process could therefore grow `file_history.db` past the configured 1 GiB safety limit indefinitely, eventually increasing SQLite work or consuming disk needed by session, memory, and jobs persistence. Live amortized enforcement restores the configured disk backstop without running stale-session GC or database-size queries on every tool write.

## Validation

- `go build ./...`
- `HOME=<temp> XDG_CONFIG_HOME=<temp>/config XDG_DATA_HOME=<temp>/data go test ./...`
- `go test -race ./internal/filetrack -count=1`
- Regression test was confirmed to fail before the fix (`368640` bytes with a `262144`-byte cap) and pass after it.
- `git diff --check`

### DIFF
--- a/internal/filetrack/store.go
+++ b/internal/filetrack/store.go
@@ -116,7 +116,7 @@ type FileDiffSide struct {
 type Options struct {
 	MaxFileBytes    int   // 0 = DefaultMaxFileBytes
 	MaxSessionBytes int   // 0 = DefaultMaxSessionBytes
-	MaxTotalBytes   int64 // 0 = DefaultMaxTotalBytes; whole-database size cap enforced by GC
+	MaxTotalBytes   int64 // 0 = DefaultMaxTotalBytes; whole-database size cap enforced live and by GC
 }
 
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #1018, opened 2026-08-09
feat: add direct shell mode to chat

sam-saffron-jarvis:
## Summary

Add Claude Code-style direct shell mode to the interactive chat composer:

- type or paste `! command` at the start of an empty composer to execute it directly
- run through the configured `$SHELL` in the session's effective directory, including bound worktrees
- stream combined stdout/stderr live in observed order
- bypass tool approval because the command is an explicit direct user action
- cancel the process group with `Esc` or `Ctrl+C`
- persist the command, directory, output, and exit state as a readable conversation turn
- automatically start the normal model response after success, nonzero exit, cancellation, or launch failure
- cap captured output at **64 KiB**, with the exact omitted byte count in the truncation marker
- recall shell history without captured output; complete partial commands from the current session with `Tab`
- keep `/shell` for interactive/full-screen PTY handoff

## Review fixes

A separate code review found and this branch fixes:

- a possible blocked completion goroutine when the TUI exits with a full event queue
- prompt-history input leaking through while a direct command owns the composer
- ambiguous/multiline shell history parsing
- paste activation of an unrelated existing draft

## Verification

- `go test ./internal/tui/chat -run 'DirectShell' -count=1`
- `go test -race ./internal/tui/chat -run 'DirectShell' -count=1`
- isolated `HOME`/XDG: `go test ./...`
- `go build ./...`
- `go vet ./...`
- `git diff --check`
- demo video rendered at 1920×1080, 30 fps and reviewed via sampled contact sheet

## Deliberate scope

This is foreground, captured shell execution. It does not add Claude's `Ctrl+B` background-task manager, and it does not pretend captured pipes are a PTY; interactive programs remain the job of `/shell`.

### DIFF
--- a/cmd/chat.go
+++ b/cmd/chat.go
@@ -101,9 +101,10 @@ Agent examples (use @agent shortcut or --agent flag):
 
 Keyboard shortcuts:
   Enter        - Send message
+  ! command    - Run directly in the session directory, then ask the model to respond
   Shift+Enter  - Insert newline
   Ctrl+/ Ctrl+H - Show help
-  Ctrl+C       - Copy selection; cancel active response/tool; press twice when idle to quit
+  Ctrl+C       - Copy selection; cancel active response/tool/shell; press twice when idle to quit
   Ctrl+K       - Clear conversation
   Ctrl+N       - New session
   Ctrl+L       - Switch model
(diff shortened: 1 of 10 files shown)


SamSaffron:
implemented on local

PR #1026, opened 2026-08-11
fix: Stalled voice transcription can exhaust every Telegram handler slot

sam-saffron-jarvis:
## What changed

- Bound Telegram voice transcription to a conservative five-minute context deadline.
- Keep `llm.TranscribeWithConfig` caller-controlled so other transcription callers retain their existing deadline policy.
- Added a per-manager timeout seam for tests.
- Added a regression test with a transcription endpoint that flushes a 200 response header and then stalls. The test fills all eight Telegram handler slots with one stalled voice request and its same-chat admission queue, then verifies the timeout cancels the provider request, releases the same-chat chain, frees every slot, and allows a different chat to execute.

## Why this is high-value

Telegram handlers retain both their global handler slot and same-chat admission ownership while voice transcription runs. The shared streaming HTTP client intentionally has no whole-request timeout, so a provider that returns headers and then stops producing its response body could previously block forever. Repeated messages from that chat could then occupy all eight global handler slots while waiting behind the first request, freezing Telegram processing for every chat until shutdown or restart.

A bounded transcription context turns that permanent bot-wide outage into the existing transcription-error reply and restores normal processing automatically.

## Validation

- `go test ./internal/serve -run 'TestTelegramVoiceTranscriptionTimeoutReleasesHandlerChain' -count=1`
- `go test -race ./internal/serve -run 'TestTelegramVoiceTranscriptionTimeoutReleasesHandlerChain' -count=1`
- `go test ./internal/serve`
- `go build ./...`
- `git diff --check`
- `go test ./...` was run; all packages except `cmd` passed. Two unrelated `cmd` tests fail because this host's 31 user skills exceed their configured visible-skill budgets, so the temporary request-directory skill is omitted: `TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` and `TestCmdRunnerPrepareUsesRequestWorkingDirForSkills`. Both failures reproduce when run individually and do not touch the Telegram changes.

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -43,6 +43,7 @@ const telegramMaxConcurrentHandlers = 8
 const telegramMaxPhotoDownloadBytes int64 = 25 << 20
 const telegramMaxVoiceDownloadBytes int64 = 25 << 20
 const telegramDownloadTimeout = 5 * time.Minute
+const telegramTranscriptionTimeout = 5 * time.Minute
 const telegramBotAPIRequestTimeout = 30 * time.Second
 const telegramUpdateLongPollSeconds = 60
 const telegramBotAPILongPollTimeout = (telegramUpdateLongPollSeconds * time.Second) + 10*time.Second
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing this review branch in favor of a fresh local integration onto current main. The Telegram transcription deadline remains applicable and is being carried forward locally.

PR #1028, opened 2026-08-12
fix: Blocked Grok ACP handshake defeats both its timeout and server shutdown

sam-saffron-jarvis:
## What changed

- Added a handshake-scoped watchdog that closes the Grok ACP process stdin and cancels its process context when startup is canceled or times out.
- Stop the watchdog before returning a successfully initialized, resumed, loaded, or newly created ACP session; if the callback already started, treat startup as failed and clean up the process.
- Added a regression test with a fake Grok process that completes initialize/authentication, then stops reading before a multi-megabyte `session/new` request. The test verifies the blocked stream, provider cleanup, and child process all unwind, then confirms a subsequent turn completes.

## Why this is high-value

ACP checks cancellation only after writing each request. A large durable-session handshake can therefore block forever in the stdin write when the Grok child remains alive but stops reading. Because durable startup holds `acpMu`, the existing deferred process stop cannot run and `CleanupMCP`/server shutdown cannot acquire the mutex to close stdin. This can permanently wedge a Jarvis conversation, leave its active-stream slot occupied, make `Stream.Close` hang, and block service shutdown. The watchdog breaks that cycle without taking the provider mutex while preserving successful handshake behavior.

## Validation

- `go test ./internal/llm -run 'TestGrokBinProviderACPCancellation(UnblocksSessionWrite|DuringInitialize|StopsProcess)$' -count=1`
- `go test ./internal/llm -run TestGrokBinProviderACPCancellationUnblocksSessionWrite -count=3`
- `go test -race ./internal/llm -run TestGrokBinProviderACPCancellationUnblocksSessionWrite -count=1`
- `go build ./...`
- `git diff --check`

### DIFF
--- a/internal/llm/grok_acp.go
+++ b/internal/llm/grok_acp.go
@@ -764,6 +764,23 @@ func (p *GrokBinProvider) startGrokACPProcess(ctx context.Context, req Request, 
 
 	handshakeCtx, handshakeCancel := context.WithTimeout(ctx, grokACPHandshakeTimeout)
 	defer handshakeCancel()
+	stopHandshakeWatchdog := context.AfterFunc(handshakeCtx, func() {
+		// ACP writes do not observe their call context while blocked in Write.
+		// Close stdin directly so startup can unwind even while acpMu is held.
+		_ = process.stdin.Close()
+		process.cancel()
+	})
+	defer stopHandshakeWatchdog()
+	finishHandshake := func() error {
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing this review branch in favor of a fresh local integration onto current main. The ACP handshake watchdog remains necessary and is being integrated with newer Grok lifecycle changes.

PR #1030, opened 2026-08-13
feat: add durable concurrent worker threads

sam-saffron-jarvis:
## Summary

Add durable concurrent worker threads to interactive chat, with an explicit mailbox boundary between worker output and coordinator context.

- `/thread TASK` starts a background worker attached to the current coordinator session
- `/tree` shows worker state and unread reports, and supports opening, cancelling/reconciling, previewing, and importing results
- `/main` returns from a worker session to its coordinator
- worker-only `report` tool emits structured `progress`, `result`, and `blocker` events
- reports remain control-plane state until explicitly imported as attributed user content
- terminal completion synthesizes a result when the worker does not emit one

## Architecture

Worker relationships and reports use dedicated session-domain tables rather than overloading conversation-fork edges:

- `session_workers` stores coordinator/child relationships and durable lifecycle state
- `session_worker_reports` stores immutable report content plus read/import metadata
- worker transcripts remain single-writer
- report bodies never receive developer/system authority

Worker execution reuses jobs-v2, with additional safety for interactive ownership:

- dedicated `thread_jobs_v2.db` isolates worker runs from ordinary and older jobs daemons
- per-run ownership leases prevent one live chat from claiming or cancelling another chat's workers
- expired ownership is reconciled without retrying side-effecting active runs
- worker approval is resolved from the durable coordinator and cannot exceed it
- every terminal path produces exactly one terminal mailbox result

The MVP deliberately keeps a flat coordinator/worker topology and does not auto-merge worktrees or steer active workers.

## Safety and recovery

- blocks report import and `/thread` creation during active transcript mutation/streaming
- reconciles stale queued/running/cancel-requested worker edges
- provides `Cancel or reconcile worker` even when startup failed before a run ID was saved
- prevents simultaneous active workers sharing a coordinator or canonicalized workspace
- keeps reports outside model context until explicit user import
- imports remain valid across compaction boundaries

## Validation

Passed:

- `go test ./...`
- `go build ./...`
- `go vet ./...`
- `go test -race ./internal/session -run 'WorkerConcurrentTerminal|WorkerPreventsConflicting'`
- `go test -race ./cmd -run 'ChatWorkerReconcile|ExpiredCancelRecovery|WorkerLoopRevisits|TwoThreadManagers'`
- `git diff --check`

Also exercised end-to-end in a virtual TTY with `chatgpt:gpt-5.6-luna`:

1. spawned a worker with `/thread`
2. observed the unread mailbox notification
3. previewed and imported its result through `/tree`
4. opened the child session and returned with `/main`
5. confirmed the coordinator model could use the imported result

## Scope

29 files, approximately `+3914/-98`. Much of the size is migrations, durable lifecycle/reconciliation, jobs ownership safety, and focused regression tests; the user-facing command surface is intentionally small.

## Interactive follow-up reporting

A resumed, completed worker can now be continued as a normal interactive session and then publish a fresh mailbox briefing with bare `/report`.

- `/report` reuses the control-plane briefing generation machinery extracted from `/handover`
- it summarizes the active post-compaction child context, including interactive follow-up turns
- it writes an immutable `result` report with `origin=interactive_report`
- it remains in the child; `/main` remains navigation only
- it never creates a normal turn or enters the coordinator context automatically
- handover-specific scripts, confirmation, target-session creation, and switching remain separate

Live virtual-TTY validation with `chatgpt:gpt-5.6-luna` covered:

1. spawn with `/thread`
2. open the completed worker
3. continue it interactively with an independent verification request
4. run bare `/report`
5. return through `/main`
6. preview/import the new interactive report
7. confirm the coordinator can answer from the imported briefing

### DIFF
--- a/cmd/chat.go
+++ b/cmd/chat.go
@@ -210,9 +210,11 @@ func runChat(cmd *cobra.Command, args []string) error {
 
 	handoverAutoSend := ""
 	relaunchHandoff := chatRelaunchHandoff{}
+	workerRuntime := &chatWorkerRuntime{}
+	defer func() { _ = workerRuntime.Close() }()
 	chatHandoverApprovalMode = nil
 	for {
-		nextResumeID, nextAutoSend, err := runChatOnce(ctx, cmd, initialText, cliAgent, resumeRequested, resumeID, handoverAutoSend, &relaunchHandoff)
+		nextResumeID, nextAutoSend, err := runChatOnce(ctx, cmd, initialText, cliAgent, resumeRequested, resumeID, handoverAutoSend, &relaunchHandoff, workerRuntime)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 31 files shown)


SamSaffron:
Closing this large review branch in favor of a fresh local integration onto current main. The durable worker-thread design remains useful; its hardened changesets are being carried forward together for current-main validation and improvement.

PR #1031, opened 2026-08-13
fix: Idle runtime cleanup can permanently wedge the session janitor

sam-saffron-jarvis:
## What changed

- Added a two-second bounded retirement path for session runtimes.
- Kept `onEvict` synchronous, then used `serveRuntime.CloseContext` so a stalled MCP `session.Close` or provider `CleanupMCP` can continue only in the runtime's isolated cleanup goroutine instead of blocking the caller forever.
- Routed TTL eviction, capacity eviction, and other manager-owned idle-runtime retirement paths through the same helper.
- Added regression coverage with a fake provider whose cleanup blocks, including both janitor progress and capacity replacement.

## Why this is high-value

Jarvis's long-running web service exercises session retirement every day. Previously, one third-party cleanup hook that never returned could permanently stop the sole TTL janitor. The same failure at the 1000-session capacity boundary could also hang the request opening a replacement session. Bounding retirement keeps reclamation and new-session requests moving while preserving eviction callbacks and eventual cleanup behavior.

## Validation

- `go build ./...`
- `go test ./cmd -run 'TestServeSessionManager_(EvictExpiredBoundsBlockingCleanup|GetOrCreateBoundsCapacityCleanup)$' -count=10`
- `go test -race ./cmd -run 'TestServeSessionManager_(EvictExpiredBoundsBlockingCleanup|GetOrCreateBoundsCapacityCleanup)$' -count=1`
- `XDG_CONFIG_HOME=<empty temp dir> go test ./...`

The initial unisolated `go test ./...` run hit two unrelated existing `cmd` test failures because Jarvis's real user skill catalogue exhausted those tests' visible-skill limits and hid their project fixtures. Re-running the full suite with an empty temporary `XDG_CONFIG_HOME` passed.

### DIFF
--- a/cmd/serve_session.go
+++ b/cmd/serve_session.go
@@ -7,11 +7,14 @@ import (
 	"time"
 )
 
+const defaultServeSessionRetirementTimeout = 2 * time.Second
+
 type serveSessionManager struct {
-	ttl     time.Duration
-	max     int
-	factory func(context.Context) (*serveRuntime, error)
-	onEvict func(rt *serveRuntime) // called when a session is evicted
+	ttl               time.Duration
+	max               int
+	factory           func(context.Context) (*serveRuntime, error)
+	onEvict           func(rt *serveRuntime) // called when a session is evicted
+	retirementTimeout time.Duration
 
 	mu       sync.Mutex
 	sessions map[string]*serveRuntime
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing this review branch in favor of a fresh local integration onto current main. The bounded runtime-retirement fix remains applicable and is being carried forward locally.

PR #1032, opened 2026-08-13
fix: An oversized CLI diagnostic line can deadlock provider streaming

sam-saffron-jarvis:
## What changed

- Replaced the 1 MiB-limited stderr scanners used by Claude, Grok, Cursor, Agy, and Grok ACP with a shared bounded line reader that continues draining oversized records.
- Keeps diagnostic memory bounded at 1 MiB per line and records an omission marker instead of retaining a partial oversized line that could bypass whole-value redaction.
- Cancels CLI subprocesses when their stdout framing scanner fails, so unread stdout cannot fill the pipe while the caller waits for process exit.
- Added fake-Claude regression coverage for both a 3 MiB unterminated stderr line and an oversized unterminated stdout record, plus direct coverage of bounded diagnostic draining.

## Why this is high-value

Jarvis uses these CLI providers on ordinary web, Telegram, and terminal turns, which do not necessarily have deadlines. Previously, one large provider diagnostic could stop the stderr scanner, fill the OS pipe, and leave the provider turn blocked forever with no final error or completion event. The same pipe-deadlock pattern existed across multiple CLI providers, so a shared draining reader removes a catastrophic stuck-turn failure from heavily used paths while preserving bounded diagnostics.

## Validation

- `gofmt -w` on all changed Go files
- `go build ./...` — passed
- `go test ./internal/llm -run 'TestDrainCLIDiagnosticLines|TestClaudeBinProvider_Oversized' -count=1` — passed
- `go test -race ./internal/llm -run 'TestDrainCLIDiagnosticLines|TestClaudeBinProvider_Oversized' -count=1` — passed
- `go test ./...` — all packages passed except two unrelated `cmd` tests (`TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` and `TestCmdRunnerPrepareUsesRequestWorkingDirForSkills`). They fail reproducibly in isolation because this host exposes 31 globally configured skills while the tests' metadata limits show only 8/10, omitting their temporary fixture skill; no `cmd` or skill-discovery code is changed here.
- `git diff --check` — passed

### DIFF
--- a/internal/llm/agy_bin.go
+++ b/internal/llm/agy_bin.go
@@ -481,15 +481,13 @@ func (p *AgyBinProvider) runCommand(ctx context.Context, args []string, prompt, 
 	}()
 	go func() {
 		defer close(stderrDone)
-		sc := bufio.NewScanner(stderr)
-		sc.Buffer(make([]byte, 64<<10), 1<<20)
-		for sc.Scan() {
-			line := redact(sc.Text())
+		_ = drainCLIDiagnosticLines(stderr, func(rawLine string) {
+			line := redact(rawLine)
 			if debug {
 				fmt.Fprintf(os.Stderr, "[agy stderr] %s\n", line)
 			}
 			recordCLITailLine(&stderrMu, &stderrTail, line, agyStderrTailMaxLines)
-		}
+		})
 	}()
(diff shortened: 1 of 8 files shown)


SamSaffron:
Closing this review branch in favor of a fresh local integration onto current main. The fix is still applicable and its changeset is being carried forward with the related Grok lifecycle hardening.

PR #1038, opened 2026-08-16
fix: Scheduled run insertion is not atomic with schedule advancement

sam-saffron-jarvis:
## What changed

- Run the concurrency count, queued-run insert, and schedule bookkeeping inside one SQLite transaction.
- Pass that transaction to the limit and post-insert callbacks, committing only after the once-job disable or cron advancement succeeds.
- Keep `enqueueMu` to preserve same-process admission ordering.
- Add a regression test that injects a schedule-advancement failure after the run insert, reopens the database, and verifies the insert rolled back before a later schedule pass queues exactly one run and disables the once job.

## Why this is high-value

Scheduled jobs are a daily Jarvis path and can execute user-visible notifications, agent work, programs, and token-consuming provider calls. Previously, a crash or database error after the run insert but before schedule advancement could leave both a durable queued run and a still-due job, allowing one-shot or cron side effects to run twice. Making those writes crash-atomic removes that duplicate-execution window with a narrowly scoped transaction.

## Validation

- `go test ./cmd -run '^(TestJobsV2ScheduledEnqueueRollsBackWhenScheduleAdvancementFails|TestJobsV2EnqueueRunWithConcurrencyLimitSerializesConcurrentEnqueues|TestJobsV2ScheduleOneRespectsMaxConcurrentRuns)$' -count=1`
- `go build ./...`
- `HOME="$TMPHOME" XDG_CONFIG_HOME="$TMPHOME/.config" GOMODCACHE=/home/agent/go/pkg/mod GOCACHE=/home/agent/.cache/go-build go test ./...`
- `git diff --check`

The regression test failed before the fix with one committed run after the injected advancement error. An initial unisolated `go test ./...` encountered two unrelated skill-discovery tests affected by the host's global skill catalog; both those tests and the full suite pass with an isolated home/config environment.

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -1283,33 +1283,46 @@ func (m *jobsV2Manager) countActiveRuns(jobID string) (int, error) {
 	return active, nil
 }
 
-func (m *jobsV2Manager) enqueueRunWithConcurrencyLimit(job jobsV2Job, trigger string, scheduledFor time.Time, onLimit func() error, afterInsert func(string) error) (string, int, bool, error) {
+func (m *jobsV2Manager) enqueueRunWithConcurrencyLimit(job jobsV2Job, trigger string, scheduledFor time.Time, onLimit func(*sql.Tx) error, afterInsert func(*sql.Tx, string) error) (string, int, bool, error) {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #1040, opened 2026-08-17
fix: Concurrent worktree merges can erase a successful merge

sam-saffron-jarvis:
## What changed

- Added a canonical per-repository-root mutation lock shared by merge-back, promote-to-root, and assisted-merge operations. The lock covers the root cleanliness check through apply, commit, conflict cleanup, or promotion rollback.
- Made cherry-pick cleanup skip `git reset --merge` unless Git has either cherry-pick metadata or unmerged index entries, so cleanup cannot reset unrelated staged changes when the failed command established no cherry-pick state.
- Added regression coverage for overlapping conflicting merge-back requests and for preserving staged changes when no cherry-pick state exists.

## Why this is high-value

Two overlapping web merge/promote requests could both observe a clean root. One request could apply successfully, then the other could conflict and unconditionally reset the first request's staged result while the first client still received `Applied=true`. A double submission or requests from multiple browser tabs could therefore silently erase a successful merge. Per-root serialization makes the later request observe the staged root and return `ErrRootDirty` instead.

## Validation

- `gofmt -w internal/worktree/worktree.go internal/worktree/worktree_test.go`
- `go test ./internal/worktree`
- `go test -race ./internal/worktree`
- `go build ./...`
- `go test ./...` with a clean `HOME`/`XDG_CONFIG_HOME` (passes; the host-configured run independently failed two existing skill-discovery tests because the host's 31 skills exceeded their visible-skill budget)
- `git diff --check`

### DIFF
--- a/internal/worktree/worktree.go
+++ b/internal/worktree/worktree.go
@@ -231,7 +231,16 @@ var (
 	ErrConflict           = errors.New("merge back has conflicts")
 	ErrRootDirty          = errors.New("root checkout has uncommitted changes")
 	ErrMergeCleanupFailed = errors.New("merge succeeded but worktree cleanup failed")
+	rootMutationLocks     sync.Map
 )
+
+func lockRootMutation(root string) func() {
+	key, _ := samePathKey(root)
+	value, _ := rootMutationLocks.LoadOrStore(key, &sync.Mutex{})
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #1041, opened 2026-08-17
fix: Root checkout mutations can race with newly admitted or non-web agent runs

sam-saffron-jarvis:
## What changed

- Added a process-wide, per-repository root checkout lease.
- Root-bound web, Telegram, jobs, and child-agent runs acquire a shared lease during admission and hold it through run completion.
- Linked-worktree runs remain independent and do not acquire the main checkout lease.
- Worktree merge and promote handlers atomically acquire an exclusive lease before checking legacy web runtime activity, and hold it through the Git operation and cleanup.
- Preserved HTTP 409 behavior when a root run or another root mutation is already active.

## Why this is high-value

Merge and promote mutate the shared root checkout's HEAD, index, and working files. Previously, their one-time web-session snapshot allowed a new web run to start after the check, and did not represent Telegram or jobs runs at all. That could make an agent read, edit, or commit while its checkout changed underneath it, mixing changes or placing commits on the wrong branch.

The shared admission lease closes both races while leaving worktree-bound agents concurrent with root mutation operations.

## Validation

- `go test -race ./cmd -run 'TestServeWorktree(MergeBlocksActiveRootRun|MutationsBlockNonWebRootRunLease|MergeExclusiveLeaseBlocksNewRootRuns)$'`
- `go build ./...`
- `HOME=<temp> XDG_CONFIG_HOME=<temp>/config XDG_DATA_HOME=<temp>/data XDG_CACHE_HOME=<temp>/cache go test ./...`
- `go vet ./...`
- `git diff --check`

Regression coverage verifies that:

- already-admitted non-web root runs make both merge and promote return HTTP 409;
- once merge has exclusive admission, web, Telegram, and jobs runtime admissions cannot become active until the Git operation releases the lease;
- the existing active web-runtime conflict behavior remains intact.

### DIFF
--- a/cmd/root_checkout_lease.go
+++ b/cmd/root_checkout_lease.go
@@ -1,0 +1,168 @@
+package cmd
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"strings"
+	"sync"
+
+	"github.com/samsaffron/term-llm/internal/worktree"
+)
+
+// rootCheckoutLeaseRegistry coordinates agent runs with operations that mutate
+// a repository's main checkout. Runs in linked worktrees use their own checkout
+// and intentionally do not participate.
+type rootCheckoutLeaseRegistry struct {
+	mu     sync.Mutex
+	leases map[string]*rootCheckoutLease
+}
+
+type rootCheckoutLease struct {
+	mu      sync.Mutex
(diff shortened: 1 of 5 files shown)


SamSaffron:

PR #1043, opened 2026-08-18
fix: Responses WebSocket silently drops the requested priority service tier

sam-saffron-jarvis:
## What changed

- Added `service_tier` to the Responses WebSocket `response.create` wire request.
- Copied `ResponsesRequest.ServiceTier` into that wire request while retaining `omitempty` behavior for an explicitly cleared or unset tier.
- Added WebSocket-frame coverage for `service_tier: "priority"` and omission coverage for an empty tier.

## Why this is high-value

ChatGPT enables the Responses WebSocket transport by default. Provider defaults and TUI `/fast` requests already normalize fast mode to the API value `priority`, but the WebSocket conversion silently discarded that value. As a result, the backend used its default queue even while the UI and configuration reported fast mode.

Forwarding the requested tier lets ChatGPT actually provide the lower-queue-latency priority service selected by the user. The existing HTTP provider tests cover both request-level and provider-default tier selection; the new wire test covers the previously missing WebSocket serialization step.

## Validation

- Reproduced before the fix: `TestResponsesClientStreamWebSocket` decoded no `service_tier` from the first `response.create` frame.
- `go test ./internal/llm -run 'TestResponses(WebSocketRequestOmitsTransportFields|ClientStreamWebSocket)$' -count=1`
- `go build ./...`
- `go test ./...` (all packages, including `internal/llm`, passed except two unrelated `cmd` tests whose host skill-catalog truncation omitted their temporary test skill: `TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` and `TestCmdRunnerPrepareUsesRequestWorkingDirForSkills`)
- `git diff --check`

### DIFF
--- a/internal/llm/responses_websocket.go
+++ b/internal/llm/responses_websocket.go
@@ -58,6 +58,7 @@ type responsesWSRequest struct {
 	Generate           *bool                        `json:"generate,omitempty"`
 	StreamOptions      *ResponsesStreamOptions      `json:"stream_options,omitempty"`
 	PreviousResponseID string                       `json:"previous_response_id,omitempty"`
+	ServiceTier        string                       `json:"service_tier,omitempty"`
 }
 
 func newResponsesWSRequest(req ResponsesRequest) responsesWSRequest {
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #1044, opened 2026-08-18
fix: Server ping heartbeats cannot keep Responses WebSocket streams alive

sam-saffron-jarvis:
## What changed

- Wrap the Gorilla WebSocket connection's existing ping handler when a Responses socket is created.
- Refresh the stream read deadline for each server ping, then delegate to Gorilla's handler so the matching pong is still sent.
- Add an integration test that keeps a stream idle for three timeout periods using server ping frames before sending `response.completed`.

## Why this is high-value

ChatGPT enables Responses-over-WebSocket by default, and long reasoning pauses are normal in Sam/Jarvis's daily usage. Gorilla's default incoming-ping handler sends a pong but does not invoke the pong handler, so server heartbeats previously could not refresh term-llm's five-minute read deadline. A healthy, heartbeat-responsive stream could therefore be discarded at the idle boundary after partial output, where automatic HTTP fallback is unsafe.

This preserves the existing idle timeout for genuinely silent connections while allowing valid server heartbeats to keep active streams alive.

## Validation

- Regression reproduced before the fix: the new test failed at the 100 ms idle deadline with `Responses WebSocket closed before response.completed: ... i/o timeout`.
- `go test ./internal/llm -run '^(TestResponsesClientStreamWebSocket|TestResponsesClientWebSocketServerPingsKeepStreamAlive|TestResponsesClientWebSocketContextCancel)$' -count=5`
- `go test -race ./internal/llm -run '^TestResponsesClientWebSocketServerPingsKeepStreamAlive$' -count=1`
- `go build ./...`
- `go test ./...` ran all packages; the changed `internal/llm` package passed. The overall command remains red on two unrelated `cmd` tests (`TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` and `TestCmdRunnerPrepareUsesRequestWorkingDirForSkills`) because this agent's 31 configured skills exceed those tests' visible-skill budgets, omitting their temporary fixture skill.
- `git diff --check`

### DIFF
--- a/internal/llm/responses_websocket.go
+++ b/internal/llm/responses_websocket.go
@@ -157,6 +157,15 @@ func (c *ResponsesClient) streamWebSocketPrepared(ctx context.Context, req Respo
 		if idleTimeout == 0 {
 			idleTimeout = 5 * time.Minute
 		}
+		if !reused {
+			pingHandler := conn.PingHandler()
+			conn.SetPingHandler(func(appData string) error {
+				if err := conn.SetReadDeadline(time.Now().Add(idleTimeout)); err != nil {
+					return err
+				}
+				return pingHandler(appData)
+			})
+		}
 		conn.SetPongHandler(func(string) error {
 			return conn.SetReadDeadline(time.Now().Add(idleTimeout))
 		})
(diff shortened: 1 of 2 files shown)


SamSaffron:

PR #1045, opened 2026-08-19
fix: One unread reverse response can wedge every request on the node tunnel

sam-saffron-jarvis:
## What changed

- Made reverse-response delivery to each request queue nonblocking so the tunnel's sole WebSocket reader cannot wait behind one unread response.
- When a request's 16-frame queue is full, remove and abort only that pending request with a slow-consumer error and asynchronously send its cancel frame to the node.
- Replaced the old node-wide backpressure expectation with an integration test that overflows an unread response, verifies a sibling request completes on the same tunnel, and checks that the unread body receives the overflow error.

## Why this is high-value

Every multiplexed request for a reverse-connected node shares one WebSocket reader. Previously, one suspended or abandoned consumer could fill its 16-frame queue (about 512 KiB with 32 KiB body chunks) and block that reader indefinitely. That prevented sibling web, health, streaming, and delegation responses from being dispatched, and also prevented pong processing until the whole node appeared offline.

Bounding the failure to the request that stopped draining preserves service for every other request and keeps the healthy reverse tunnel connected.

## Validation

- Reproduced the failure with the new integration test before the implementation: it timed out waiting for cancellation because the shared reader was wedged.
- `go build ./...`
- `go test ./cmd -run 'TestHubReverse' -count=1`
- `go test -race ./cmd -run '^TestHubReverse(ReadLoopIsolatesUnreadResponseOverflow|ReadLoopSurvivesCanceledUndrainedStream|ResponseCancellation)$' -count=1`
- `go test ./...` *(all non-`cmd` packages passed; `cmd` has two unrelated ambient-skill-discovery failures because the host's visible-skill budget omits temporary test skills: `TestResolveChatRuntimeSystemContextUsesExplicitDirectoryWithoutChdir` and `TestCmdRunnerPrepareUsesRequestWorkingDirForSkills`)*
- `git diff --stat`
- `git diff --check`

### DIFF
--- a/cmd/serve_hub_reverse.go
+++ b/cmd/serve_hub_reverse.go
@@ -39,6 +39,8 @@ const (
 	hubReverseFrameResponseEnd   = "response_end"
 )
 
+var errHubReverseSlowConsumer = errors.New("reverse response consumer is too slow")
+
 type hubReverseRequest struct {
 	Type   string      `json:"type,omitempty"`
 	ID     string      `json:"id"`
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing this PR as requested. The important reverse-tunnel slow-consumer isolation fix and regression coverage will be implemented directly in the current working tree without committing.

PR #1046, opened 2026-08-19
fix: Every stateful response run decodes all historical transcript bodies before provider start

sam-saffron-jarvis:
## What changed

- Added an optional response-run start-state reader for session stores.
- Implemented the SQLite reader as one coherent scalar query for `transcript_rev`, compaction sequence/count, and the latest eligible durable boundary row ID.
- Routed the capability through `LoggingStore` and made `configureResponseRunRevision` prefer it while retaining the existing full-snapshot path for custom, unversioned, or otherwise unsupported stores.
- Added parity coverage across compaction-tail and mixed-role rows, a serve-level test proving the logged SQLite path avoids `GetTranscriptSnapshot`, and a large-body benchmark.

## Why this is high-value

Every genuinely new stateful UI, skill, or queued-job continuation configures its response run before the provider starts and before `response.created` is emitted. The previous path selected every historical `parts` blob, decoded every blob into `[]llm.Part`, retained transcript-sized slices, and built display/tool-call state even though response startup only needs four scalar values.

The SQLite fast path now reads no historical body bytes and uses the existing session/sequence indexes to find the newest eligible boundary. It also uses the primary SQLite connection, so startup is not queued behind a long body materialization on the dedicated transcript reader. Custom stores preserve existing behavior through the snapshot fallback.

On an i9-14900K with alternating 32 KiB tool results and inline image payloads, the committed benchmark showed the compact configure path remaining about 20 microseconds and 3.5 KiB allocated from 1 to 256 rows. The full-snapshot reference grew from about 0.19 ms / 159 KiB / 85 allocations to 35–39 ms / 40.1 MiB / 7,016 allocations at 256 rows.

## Validation

- `go test ./internal/session -run 'TestSQLiteResponseRunStartStateMatchesTranscriptSnapshot' -count=1`
- `go test ./cmd -run 'TestConfigureResponseRunRevision(UsesCompactStartStateThroughLoggingStore|SkipsTrailingNonBranchableRows|BoundsStartedRevisionRead)$' -count=1`
- `go test ./cmd -run '^$' -bench '^BenchmarkConfigureResponseRunRevisionLargeBodies$' -benchmem -benchtime=300ms -count=3`
- `go build ./...`
- `go test ./...` with isolated `HOME`/XDG directories as documented in `AGENTS.md`
- `go vet ./...`
- `git diff --check`

### DIFF
--- a/cmd/serve_response_runs.go
+++ b/cmd/serve_response_runs.go
@@ -2590,24 +2590,38 @@ func (s *serveServer) configureResponseRunRevision(run *responseRun, sessionID s
 		return
 	}
 	startedCtx, startedCancel := context.WithTimeout(context.Background(), responseRunRevisionReadTimeout)
-	if indexer, ok := s.transcriptIndexerForWeb(); ok {
-		if snapshot, err := indexer.GetTranscriptSnapshot(startedCtx, sessionID); err == nil {
-			run.startedRev = snapshot.Rev
-			run.startedCompactionSeq = snapshot.CompactionSeq
-			run.startedCompactionCount = snapshot.CompactionCount
-			for i := len(snapshot.Items) - 1; i >= 0; i-- {
-				item := snapshot.Items[i]
-				if item.ID <= 0 || item.Flags&session.TranscriptFlagCompactionTail != 0 {
-					continue
-				}
-				switch llm.Role(item.Role) {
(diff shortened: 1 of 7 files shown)


SamSaffron:
Closing this PR as requested. The important compact response-run start-state path and essential regression coverage will be implemented directly in the current working tree without committing; the optional benchmark will be skipped.

PR #1050, opened 2026-08-20
Expand inline diff review workflow

sam-saffron-jarvis:
## Summary

Build out the inline diff review loop after the initial line-comment feature:

- preserve focus when responses land and across diff body/list reorders
- pin commented files open so live-follow cannot collapse or scroll away from the review target
- replace the text-heavy bulk control with an icon and add a same-DOM full-window Changes view
- add session-scoped queued inline comments and deliver a review batch as one user message / one agent run

## Review flow

The inline editor now has a split action:

- **Send now** sends only the current anchored instruction
- **Queue comment** captures the path, side, line, sequence, exact text, context, and instruction without starting a run

The sticky queue bar sends N queued comments as exactly one request carrying N `diff_comment` parts and one provider-facing text block. Direct sends never silently absorb the queue. Failed or branch-context-parked sends preserve authored comments and retry identity.

Queue limits are **20 comments client-side**, **25 `diff_comment` parts server-side**, and **256 KiB aggregate metadata** per message.

## Changes surface

- icon-only expand/collapse with dynamic accessible labels
- maximize/restore covers the window without reparenting or rebuilding the panel
- expansion state, scroll anchor, editor drafts, and focus survive maximize/restore
- main/sidebar become inert while maximized; mobile drawer mode omits maximize
- Escape closes the innermost menu/editor first, then restores the panel, then closes the drawer

## Reliability and accessibility

- response completion restores composer focus only when focus is still idle
- body rebuilds and file-order changes restore the exact focused comment control only when it still exists
- commenting promotes an auto-expanded file to explicit `userExpanded`
- stale/removed panels no longer suppress live-follow forever
- split menu has complete focus, arrow-key, Home/End, Tab, outside-pointer, and Escape behavior
- persistent polite queue status, deliberate focus before queue bar hides, and disabled mutations during send
- queued state uses dedicated storage, not the sent-intent registry
- plural `diffComments` flows through send, interjection, deferred retry, sanitize, conversion, and transcript rendering while retaining singular compatibility

## Verification

- every `internal/serveui/static/*_test.js` suite
- `node --check` for all first-party static JavaScript
- isolated `HOME`/XDG `go test ./...`
- `go build ./...`
- `go vet ./...`
- `git diff --check`
- live Chromium run with three real Ruby files: queue three anchored comments, send one batch, apply all three changes, keep typing in an unrelated inline draft while the response completes, verify focus remains in the textarea, inspect persisted sent markers, and restore from maximized Changes

## Review process

- Fable reviewed and refined the product interaction plan
- Opus reviewed the full implementation
- all blocker/major findings and robustness suggestions were addressed
- the live demo exposed two additional real-browser focus-loss paths (detached body rebuild and list reorder); both now have realistic regression tests

### DIFF
--- a/cmd/serve_diff_comments_test.go
+++ b/cmd/serve_diff_comments_test.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"net/http/httptest"
 	"path/filepath"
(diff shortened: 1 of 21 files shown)


SamSaffron:
already merged

PR #1053, opened 2026-08-22
fix: Script-backed custom tools report execution failures as successes

sam-saffron-jarvis:
## What changed

- Added a shared script-tool error-output helper that preserves existing formatted diagnostics while setting `ToolOutput.IsError=true` for validation, resolution, setup, and execution failures.
- Marked captured nonzero exit statuses as errors in both `CustomScriptTool` and `RunAgentScriptTool`.
- Added regression coverage for missing scripts, unexecutable scripts, diagnostic output followed by exit 7, validation failures, and successful execution.

## Why this is high-value

Jarvis uses agent- and skill-declared script tools for routine memory, notification, infrastructure, and integration work. These tools previously returned failure text and nonzero exit codes with a successful status, so the engine could emit `ToolSuccess=true`, persist `ToolResult.IsError=false`, and allow the model to treat an operation that never completed as successful. Correct failure metadata now reaches every caller without changing script output formatting or timeout handling.

## Validation

- `gofmt -w internal/tools/custom_tool.go internal/tools/custom_tool_test.go internal/tools/run_agent_script.go internal/tools/run_agent_script_test.go`
- `go build ./...`
- `go test ./internal/tools -run 'Test(CustomScriptTool|RunAgentScriptTool)' -count=1`
- `go test ./internal/llm -run 'TestEngineToolOutputIsErrorMarksToolExecEndFailed|TestEngineTimeoutOutputContinuesLoop' -count=1`
- `go test ./...` reached and passed the changed `internal/tools` package, but the full run was blocked by unrelated failures in untouched packages: two `cmd` skill-discovery tests see the host's 31 configured skills instead of their request-directory fixture within the configured visibility budget, and `internal/filetrack.TestRecordChangeEnforcesTotalBudget` reports over-pruning.
- `git diff --check`

### DIFF
--- a/internal/tools/custom_tool.go
+++ b/internal/tools/custom_tool.go
@@ -29,6 +29,12 @@ type CustomScriptTool struct {
 	toolConfig *ToolConfig
 }
 
+func scriptToolErrorOutput(err *ToolError) llm.ToolOutput {
+	output := llm.TextOutput(formatToolError(err))
+	output.IsError = true
+	return output
+}
+
 // newCustomScriptTool creates a CustomScriptTool from a definition and agent directory.
 func newCustomScriptTool(def agents.CustomToolDef, agentDir string, limits OutputLimits, toolConfigs ...*ToolConfig) *CustomScriptTool {
 	return &CustomScriptTool{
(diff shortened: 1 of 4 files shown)


SamSaffron:
Closing this PR as requested. The important script-tool failure classification fix and regression coverage will be implemented directly in the current working tree without committing.

PR #1054, opened 2026-08-22
fix: CLI-provider MCP bridge discards tool failure status

sam-saffron-jarvis:
## What changed

- Extended the internal MCP HTTP executor result to carry semantic tool failure status separately from Go transport errors.
- Propagated `ToolOutput.IsError || ToolOutput.TimedOut` through the shared CLI tool bridge and into `mcp.CallToolResult.IsError`.
- Kept the standalone `serve mcp` executor consistent with the new result shape.
- Added an end-to-end HTTP MCP regression test covering success, semantic failure, timeout, and non-nil Go error responses through the bridge-backed executor.

## Why this is high-value

Claude Bin, Grok Bin, Cursor Bin, and Agy Bin all use this shared MCP bridge. Tools such as `shell` intentionally return a nil Go error with `ToolOutput.IsError=true` for nonzero exits, but the old string-only adapter discarded that status and reported the MCP call as successful. That could make ordinary Jarvis runs treat failed file, service, or notification work as completed and skip retry or recovery.

The bridge now preserves the engine's existing failure classification without changing tool output text or transport-error behavior.

## Validation

- `go test ./internal/llm ./internal/mcphttp`
- `go build ./...`
- `go vet ./...`
- `go test ./...` was run both directly and with an isolated HOME. The changed packages passed; the full suite remains blocked by the unrelated existing `internal/filetrack.TestRecordChangeEnforcesTotalBudget` failure (`live budget enforcement pruned more sessions than required`). The direct run also hit user-home skill-discovery assumptions in two `cmd` tests, which passed with the repository-recommended isolated HOME.

### DIFF
--- a/cmd/serve_mcp.go
+++ b/cmd/serve_mcp.go
@@ -265,33 +265,33 @@ func runServeMCP(cmd *cobra.Command, args []string) error {
 	}
 
 	// Build executor that routes to the right tool.
-	executor := func(ctx context.Context, name string, args json.RawMessage) (string, error) {
+	executor := func(ctx context.Context, name string, args json.RawMessage) (mcphttp.ToolResult, error) {
 		// Check web tools first.
 		if name == mcpWebSearchToolName && webSearchTool != nil {
 			out, err := webSearchTool.Execute(ctx, args)
 			if err != nil {
-				return "", err
+				return mcphttp.ToolResult{}, err
 			}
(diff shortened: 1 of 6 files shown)


SamSaffron:
Closing this PR as requested. The important MCP failure-status propagation fix and regression coverage will be implemented directly in the current working tree without committing.

PR #1058, opened 2026-08-23
fix: update-recent rescans and reloads every completed session on every run

sam-saffron-jarvis:
## What changed

- Use the persisted update-recent checkpoint as an inclusive `updated_at` candidate boundary when listing completed sessions.
- Push agent and update-time filtering into `session.Store.List`, expose `Agent` on `SessionSummary`, and remove the per-summary `Get` lookup.
- Keep per-session offsets as the exact idempotency boundary. Capture the next checkpoint before enumeration and only advance it after the candidate set is exhausted, so concurrent activity and input-cap backlogs remain eligible.
- Add counting-store coverage with 1,000 historical sessions plus SQLite coverage for agent/time filtering and inclusive timestamp ties.

## Why this is high-value

`memory update-recent` is a recurring Jarvis maintenance job. Previously, even a no-op run paged through every completed session, loaded each session to discover its agent, read every stored offset, and queried every matching transcript tail. That work and SQLite contention grew with the entire durable session history.

After this change, steady-state no-op work is bounded to the current-session lookup and one indexed candidate-list query. Historical sessions older than the successful checkpoint produce no `Get`, offset, or `GetMessagesFrom` calls. The existing `idx_sessions_updated_at` index supplies the timestamp range scan; agent filtering is applied before rows leave SQLite.

## Validation

- `go test ./cmd -run 'TestCollectMemoryUpdateRecentInput|TestReadLastUpdatedRecentAt|TestReadWriteUpdateRecentOffset' -count=1`
- `go test ./internal/session -run 'TestSQLiteStoreListFiltersByAgentAndUpdatedAt|TestSQLiteStoreListByNumberCursor' -count=1`
- `go build ./...`
- `go vet ./cmd ./internal/session`
- `git diff --check`
- `go test ./...` ran; all touched packages passed. The repository-wide run remains blocked by pre-existing `internal/filetrack.TestRecordChangeEnforcesTotalBudget` (`live budget enforcement pruned more sessions than required`), which reproduces alone and has no diff on this branch.

### DIFF
--- a/cmd/memory_freshen.go
+++ b/cmd/memory_freshen.go
@@ -121,7 +121,11 @@ func runMemoryUpdateRecent(cmd *cobra.Command, args []string) error {
 	}
 
 	if memoryDryRun {
-		input, err := collectMemoryUpdateRecentInput(ctx, store, sessStore, agentName)
+		lastUpdatedRecentAt, err := readLastUpdatedRecentAt(ctx, store, agentName)
+		if err != nil {
+			return err
+		}
+		input, err := collectMemoryUpdateRecentInput(ctx, store, sessStore, agentName, lastUpdatedRecentAt)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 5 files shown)


SamSaffron:
Brought this change onto local main and superseded the PR implementation locally. The local version keeps the agent-filtering/no-Get optimization while avoiding timezone-unsafe SQLite timestamp comparison, uses a migration-safe exhausted-run checkpoint, and adds cap/checkpoint regressions.

PR #1060, opened 2026-08-24
fix: Isolated skill completion can persist a visible result without the context the parent model needs

sam-saffron-jarvis:
## What changed

- Added an atomic `AddMessages` session-store operation and implemented it as one SQLite transaction with ordered sequence allocation.
- Persist isolated-skill completion events and their developer-context messages as one batch using the boundary context.
- Update the active runtime history only after the full batch commits.
- Added regression coverage for a failure on the second insert, rollback, and a successful retry that commits both messages in order and appends exactly one developer result.

## Why this is high-value

Isolated skills are part of Jarvis's normal review, search, and research path. Previously, the visible completion event could commit before the developer message failed or the process exited. That left the transcript advertising a successful child run while the parent model silently lost the completed output, often forcing expensive work to be rerun. The transaction removes that split-write window without changing successful behavior.

## Validation

- `gofmt -w cmd/serve_skills.go cmd/serve_skills_test.go cmd/serve_runtime_test.go internal/session/store.go internal/session/sqlite.go internal/session/sqlite_test.go internal/session/logger.go internal/session/noop.go`
- `go build ./...`
- `go test ./internal/session -run 'TestSQLiteStoreAddMessagesRollsBackOnSecondInsertFailure' -count=1`
- `go test ./cmd -run 'TestServeSkillRunResult(BatchFailureLeavesNoVisibleHalfCompletion|SerializesWithRuntimeAndUpdatesHistory|WaitsForParentTurnBoundary)' -count=1`
- `go test ./...` (changed packages and regression tests pass; the full run remains red on unrelated pre-existing environment-sensitive skill discovery tests in `cmd` and `TestRecordChangeEnforcesTotalBudget` in `internal/filetrack`)
- `git diff --check`

### DIFF
--- a/cmd/serve_runtime_test.go
+++ b/cmd/serve_runtime_test.go
@@ -445,6 +445,33 @@ func (s *serveRuntimeTestStore) AddMessage(ctx context.Context, sessionID string
 	return nil
 }
 
+func (s *serveRuntimeTestStore) AddMessages(ctx context.Context, sessionID string, messages []*session.Message) error {
+	if s.addMessageHook != nil {
+		for _, msg := range messages {
+			if err := s.addMessageHook(ctx, sessionID, msg); err != nil {
+				return err
+			}
+		}
+	}
+	s.mu.Lock()
+	defer s.mu.Unlock()
+	staged := make([]session.Message, len(messages))
+	nextID := s.nextID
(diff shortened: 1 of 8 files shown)


SamSaffron:

PR #1061, opened 2026-08-24
fix(serveui): restore pinned sidebar section

janzenisaac:
## Summary

- restore the global **Pinned** section above project groups
- keep pinned conversations out of their project group to avoid duplicates
- preserve activity ordering and search behavior


SamSaffron:
this is restored now :hugs:

PR #1063, opened 2026-08-25
fix: Project startup reconciliation performs a full Git workspace scan once per project

sam-saffron-jarvis:
## What changed

- Reworked `ReconcileAll` to validate active project roots once, page through unassigned sessions once, and cache one canonical identity lookup per distinct persisted workspace.
- Grouped session snapshots by canonical project identity, then retained the existing atomic `ClaimProjectSessions` snapshot recheck for each matched project.
- Factored workspace identity resolution out of `MatchesWorkspace` so startup reconciliation and individual project matching use the same Git, managed-worktree, and non-Git validation rules.
- Added a focused four-project/13-workspace test that injects the resolver and asserts exactly 13 resolver calls rather than 52.

## Why this is high-value

Project reconciliation runs on every projects-enabled web startup. Previously each active project repeated the full `NoProject` session walk and re-ran Git repository/root probes for every distinct workspace, making both SQLite paging and short-lived Git process creation scale as O(projects × workspaces). The new path performs one session walk and reduces canonicalization to O(projects + distinct workspaces), while preserving atomic assignment checks and unavailable-project behavior.

## Validation

- `go build ./...`
- `go test -race ./internal/project`
- `go test -run 'TestReconcileAllResolvesEachWorkspaceOnceAcrossProjects' -count=20 ./internal/project`
- `go test ./internal/project`
- `git diff --check`
- `go test ./...` was also run. The changed package passed; the aggregate run hit unrelated current-tree failures in `internal/filetrack` (`TestRecordChangeEnforcesTotalBudget`) and two `cmd` skill-discovery tests because the host's 31 configured skills exhausted their visibility limits. The two `cmd` tests pass with a clean `HOME`.

### DIFF
--- a/internal/project/assign.go
+++ b/internal/project/assign.go
@@ -133,6 +133,12 @@ func ClaimMatchingSessions(ctx context.Context, store session.Store, p session.P
 // currently available project. Unavailable project paths are skipped so a later
 // run can repair them when the filesystem is mounted again.
 func ReconcileAll(ctx context.Context, store session.Store) (int, error) {
+	return reconcileAll(ctx, store, resolveWorkspaceIdentity)
+}
+
+type reconciliationWorkspaceResolver func(context.Context, string, string) (workspaceIdentity, bool)
+
+func reconcileAll(ctx context.Context, store session.Store, resolveWorkspace reconciliationWorkspaceResolver) (int, error) {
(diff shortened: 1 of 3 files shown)


SamSaffron:

PR #1064, opened 2026-08-26
fix: Terminal queued-agent notifications are lost across shutdowns and crashes

sam-saffron-jarvis:
## What changed

- Added durable `notification_pending` and `notification_delivered_at` state to terminal job runs, including a jobs schema v2 migration and a partial index for startup replay.
- Set notification state in the same SQLite update that makes a run terminal, eliminating the crash window between terminal persistence and enqueueing process-local work.
- Replay pending terminal-run notifications during manager startup and clear the pending marker only after the notifier returns successfully.
- Suppress immediate dispatch while startup recovery terminalizes interrupted runs, then replay the resulting pending rows once through the common startup path.
- Added integration coverage for shutdown cancellation/reopen delivery, crash-equivalent terminal persistence before dispatch, successful delivery marking, one-time replay, and version 1 schema migration.

## Why this is high-value

Jarvis and scheduled `queue_agent` jobs commonly use `notify_when_done` to continue a web session or report back to Telegram. Previously, a normal serve reload, crash, or exhausted in-process retry window could leave the completed result only in the jobs database while the originating conversation never heard back. Persisting delivery state with the terminal transition turns these silent losses into startup-recoverable work while preserving asynchronous completion behavior.

The run ID remains the durable delivery identity, and successfully delivered rows are marked so later restarts do not replay them again.

## Validation

- `go test ./cmd -run 'TestJobsV2(Migration|Migrates|PendingNotification|CloseCancels|Notify)' -count=1`
- `go test -race ./cmd -run 'TestJobsV2PendingNotification' -count=1`
- `go build ./...`
- `git diff --check`
- `go test ./...` was run; all affected jobs tests and most packages passed, but the full command remains blocked by unrelated failures: two environment-sensitive skill-discovery tests select the configured global skills instead of their temporary fixture, and `internal/filetrack`'s `TestRecordChangeEnforcesTotalBudget` fails in the untouched package when run independently as well.

### DIFF
--- a/cmd/serve_jobs_v2.go
+++ b/cmd/serve_jobs_v2.go
@@ -600,6 +600,11 @@ func newJobsV2ManagerWithNotifier(dbPath string, workers int, llmExec serveJobsE
 		_ = db.Close()
 		return nil, err
 	}
+	if err := mgr.replayPendingRunDoneNotifications(); err != nil {
+		notifyCancel()
+		_ = db.Close()
+		return nil, err
+	}
 
 	mgr.wg.Add(1)
 	go mgr.schedulerLoop()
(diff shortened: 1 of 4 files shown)


SamSaffron:

PR #1066, opened 2026-08-28
feat(herdr): integrate chat lifecycle with Herdr

KrisKotlarek:
Report idle, working, and blocked chat states to the containing Herdr pane, including the active term-llm session identity. Keep reporting ordered and non-blocking, transfer authority across session switches, and release it on exit.

Support Herdr environments that expose the CLI only through PATH, use the pane-first command syntax accepted by Herdr 0.8.0, and avoid stale sequence state across repeated launches. Add lifecycle and reporter regression coverage plus an integration guide.

Screenshot 2026-08-28 at 12 07 28 pm


SamSaffron:
thanks merged in https://github.com/SamSaffron/term-llm/pull/1075 :hugs:

PR #1092, opened 2026-08-31
fix(web): scope interactive prompt modals to their own conversation

janzenisaac:
## Bug

If the agent in conversation 1 calls `ask_user` (including multi-select questions) while you are viewing conversation 2, the question popup opens over conversation 2. Same for approval prompts.

## Cause

In `RunEngine.applyResponseEvent`, a live `response.ask_user.prompt` / `response.approval.prompt` event calls `InteractionStore.present()`, which sets the global `askUser` / `approval` modal signals unconditionally — regardless of which session the event belongs to. Neighbouring state updates are already scoped (plan updates and the snapshot-recovery path both check `sessionId === sessionStore.activeSessionId`), but the live prompt path was not.

## Fix

- `InteractionStore.present()` takes an `open` flag; when false it only upserts the waiting interaction record without opening the modal.
- `RunEngine` passes `sessionId === activeSessionId` for both prompt kinds.

Background prompts remain tracked as waiting interaction records (visible in the interactions panel, sidebar attention), and selecting that conversation still opens them: selection sync already reopens pending prompts from server `pending_ask_users` / `pending_approvals` state.

## Repro / test

New test in `app-store.test.ts` (`keeps a background session interactive prompt from opening over the active conversation`) fails on main — applying an `ask_user` prompt event for background session `s1` while `s2` is active sets the global modal — and passes with the fix, while asserting active-session prompts still open immediately.

## Validation

- `vitest run`: 33/34 files pass; `webrtc.test.ts` fails identically on clean main (pre-existing, environment-related)
- `tsc --noEmit`, `eslint --max-warnings 0`, `prettier --check` clean on touched files

### DIFF
--- a/frontend/src/stores/app-store.test.ts
+++ b/frontend/src/stores/app-store.test.ts
@@ -5117,6 +5117,71 @@ describe('AppStore compatibility behavior', () => {
     await vi.waitFor(() => expect(store.streamResponse).toHaveBeenCalledWith('r1', 's1', 4));
   });
 
+  it('keeps a background session interactive prompt from opening over the active conversation', () => {
+    const store = new AppStore(config);
+    try {
+      store.sessions.value = [
+        { ...session(), activeResponseId: 'r1' },
+        { ...session(), id: 's2', title: 'Other', activeResponseId: 'r2' },
+      ];
+      store.activeSessionId.value = 's2';
+      store.draftActive.value = false;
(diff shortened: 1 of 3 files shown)


SamSaffron:
thanks @janzenisaac ... I think I just sorted it ... will also clean up tests next!

PR #1105, opened 2026-09-02
fix: Telegram interruption can hang forever while closing an uncooperative stream

sam-saffron-jarvis:
## What changed

- Cancel Telegram response contexts before beginning stream cleanup.
- Run `Stream.Close` asynchronously and bound cleanup across the stream consumer, direct stream close, and shared runner.
- Mark runtimes stale when cleanup misses the deadline so the next message replaces the runtime instead of reusing one still owned by detached work.
- Skip abnormal transcript reconciliation when a detached producer could still mutate callback output.
- Add table-driven regression coverage for an uncooperative `Stream.Close` on user interruption and a `Runner.Run` that ignores watchdog cancellation.

## Why this is high-value

Telegram holds the per-chat session mutex for the full response. Previously, interruption and error paths could call `Stream.Close` or wait for runner shutdown without a deadline, leaving `replyDone` open and the session mutex locked forever. A handful of stalled responses could retain every Telegram handler slot and stop the bot globally. Cleanup is now bounded, the old runtime is quarantined, and a replacement session can process the next message.

## Validation

- `go test ./internal/serve -run 'TestStreamReply_UncooperativeCleanupIsBounded|TestStreamReply_PersistsInterruptedPartialAssistantReply|TestStreamReply_WatchdogTimeoutIsNotTreatedAsUserInterrupt' -count=1`
- `go test -race ./internal/serve -run TestStreamReply_UncooperativeCleanupIsBounded -count=1`
- `go build ./...`
- `git diff --check`

### DIFF
--- a/internal/serve/telegram.go
+++ b/internal/serve/telegram.go
@@ -1843,20 +1843,45 @@ func (m *telegramSessionMgr) streamReplyWithAdmission(ctx context.Context, bot b
 
 	var stream llm.Stream
 	var runnerDone <-chan struct{}
-	waitForRunnerDone := func() {
-		if runnerDone == nil {
+	streamCloseDone := make(chan struct{})
+	var streamCloseOnce sync.Once
+	startStreamClose := func() {
+		streamCloseOnce.Do(func() {
+			if stream == nil {
+				close(streamCloseDone)
+				return
+			}
+			go func() {
+				_ = stream.Close()
+				close(streamCloseDone)
+			}()
+		})
+	}
+	cleanupTimeout := func() time.Duration {
+		timeout := telegramRunnerCleanupTimeout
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing at the maintainer’s request in favor of a consolidated implementation on the current working tree (not merged or pushed yet). Retained as a 9/10 priority, with additional hardening: a single cleanup deadline covers Close, the consumer, and the runner; incoming messages re-check quarantine after interruption; stale runtimes are rejected at stream entry. Added stuck-Recv and incoming-interrupt replacement regressions alongside the proposed stuck-Close/runner tests. Regression repeats and affected packages pass under the race detector; full Go suite, build, vet, and diff checks pass.

PR #1108, opened 2026-09-03
fix: A hung Git hook wedges commit operations and can survive server restart

sam-saffron-jarvis:
## What changed

- Bound the host-owned `git commit` subprocess to a generous five-minute deadline without making an admitted commit depend on client cancellation.
- Run Git in its own process group so deadline cancellation kills hooks and helper descendants, and add a bounded `WaitDelay` so inherited output pipes cannot keep `Wait` blocked.
- Preserve the existing post-process HEAD verification, so a timeout racing with a completed commit is still reported as success when the resulting commit can be verified.
- Add an integration test with a blocking pre-commit hook. It verifies the deadline-backed failure, hook termination, checkout lease release, and a successful follow-up commit.

## Why this is high-value

Both web and TUI commit workflows use this host-owned transaction path, and it holds the checkout mutex plus the outer mutation lease while Git runs. Previously, a malformed hook or stuck signing/helper process could wait forever, permanently blocking later commits and coordinated root mutations for that checkout and preventing clean server shutdown. The bounded process-group execution turns that indefinite checkout outage into a classified, recoverable failure while retaining commit-outcome verification.

## Validation

- `gofmt -w internal/gitcommit/process.go internal/gitcommit/gitcommit_test.go`
- `go build ./...` (passed after generating the ignored frontend assets required by `go:embed`)
- `go test ./internal/gitcommit -run 'TestCommitTimesOutHungHookAndReleasesCheckout|TestUnbornHookFailureIsRecoverableNotUncertain' -count=1`
- `go test ./internal/gitcommit -count=3`
- `go test -race ./internal/gitcommit -count=1`
- `go test ./...` in a clean user environment: every package passed except the unrelated existing `internal/serveui.TestProductionBundleSizeBudgets`; current generated `app.js`/`app.css` gzip sizes exceed main's budgets by 130/291 bytes. This change does not touch frontend sources or bundle budgets.
- `git diff --check`

### DIFF
--- a/internal/gitcommit/gitcommit_test.go
+++ b/internal/gitcommit/gitcommit_test.go
@@ -7,8 +7,12 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
+	"time"
+
+	"github.com/samsaffron/term-llm/internal/testutil"
 )
 
 func gitTest(t *testing.T, dir string, args ...string) string {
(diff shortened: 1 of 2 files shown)


SamSaffron:
Closing at the maintainer’s request in favor of a consolidated implementation on the current working tree (not merged or pushed yet). Retained as a 9/10 priority: bound host-owned commits, kill the process group on timeout, and preserve HEAD outcome verification. Adapted output handling to os/exec-managed writers so WaitDelay bounds pipe draining without the manual StdoutPipe/Wait ordering hazard. Hung-hook recovery test, package race tests, full Go suite, build, vet, and diff checks pass.

PR #1115, opened 2026-09-06
Disable approval controls for tool-less sessions

arpitjalan:
## Summary

- report whether a session runtime has approval-managed tools
- disable approval mode choices and Save for tool-less sessions
- replace the misleading Guardian warning and raw conflict with an actionable explanation
- check tool availability after materializing a blank session

## Testing

- make build
- go test ./cmd -run TestHandleSessionApprovalMode -count=1
- go test -p 1 ./internal/llm ./internal/webrtc
- npm --prefix frontend test (735 tests)
- npm --prefix frontend run format:check
- npm --prefix frontend run lint
- npm --prefix frontend run typecheck
- scripts/verify_nested_modules.sh
- VERIFY_RACE=1 scripts/verify_nested_modules.sh
- browser smoke test against an isolated tool-less local server

## Full-suite note

go test ./... is currently blocked by the existing TestProjectApprovals_StorageLocationHumanReadable timeout on macOS: its git rev-parse child hangs in the /home automount. The unrelated internal/llm and internal/webrtc timing failures from that concurrent run pass when rerun serially.

### DIFF
--- a/cmd/serve_approval_mode_handler.go
+++ b/cmd/serve_approval_mode_handler.go
@@ -18,6 +18,7 @@ func runtimeApprovalPolicy(rt *serveRuntime) map[string]any {
 		"default_mode":            tools.ModePrompt.String(),
 		"requested_mode":          tools.ModePrompt.String(),
 		"effective_mode":          tools.ModePrompt.String(),
+		"controls_available":      false,
 		"guardian_available":      false,
 		"guardian_auto_suspended": false,
 	}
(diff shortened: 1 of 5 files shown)


arpitjalan:
A bit more context:

The approvals dialog previously offered approval modes even when the conversation’s runtime had no approval-managed tools. It could show a Guardian availability warning, suggesting a configuration problem, but selecting another mode and saving still failed with “this session has no approval-managed tools.”

This change makes the dialog reflect the runtime’s actual capabilities. The backend reports whether approval controls are available, and the UI disables the mode choices and Save button when they are not. An explanation directs the user to choose a tool-enabled agent.

New conversations also check availability after their session is created, before attempting to save an approval mode. Older servers that omit the new availability field retain the existing behavior.

This corrects misleading UI and prevents a predictable failed request. It does not change tool permissions, approval enforcement, or Guardian behavior. Regression tests cover availability reporting and the disabled dialog state.

PR #1117, opened 2026-09-06
feat: make every config credential vaultable and resolve it lazily

sam-saffron-jarvis:
Prompted by [@justinmk](https://x.com/justinmk/status/2096610375902838932): *"looks like ~/.config/term-llm/config.yaml hardcodes secrets so i can't track it in public dotfiles"*.

He was right about the gap. Provider `api_key`/`url`/`base_url` and Bedrock keys already supported `op://`, `srv://`, `file://`, `$()` and `${VAR}`. Everything else did not:

- `image.*`, `audio.*`, `music.*`, `transcription.*`, `embed.*`, `search.*` ran only `expandEnv`, so `${VAR}` worked but `op://`, `file://` and `$()` silently ended up as literal API keys
- `serve.telegram.token` and `serve.web_push.vapid_*` supported nothing
- `mcp.json` `headers`, `env` had to hold literal tokens

## What changed

**`config.CredentialRef`** (`internal/config/credentials.go`): a config value plus its environment fallbacks, with two distinct operations:

- `Configured()` — is a value available? Never reads a vault, file, DNS or subprocess.
- `Resolve()` — actually resolve it, memoized per process so a long-running `serve` unlocks each secret at most once. Failures are not cached, so a locked vault can be retried after signing in.

Every sensitive key in `config.yaml` now accepts the full deferred syntax, plus `search.google.cx`, `search.exa_mcp.url` and `embed.ollama.base_url`. `mcp.json` gains deferred `headers`, `env` and a new `oauth.client_secret` (`client_secret_env` still works). In `mcp.json` maps only explicitly deferred forms are resolved, so a literal value starting with `$` is passed through unchanged.

## Laziness is the point

Config load no longer resolves feature credentials at all. Resolution moved to the point of use:

| Surface | Resolves when |
| --- | --- |
| image / audio / music / video / transcription | the provider is constructed for a generation |
| search | first query (`NewLazySearcher`, lazy Exa MCP client) |
| search tool advertising | never — `search.Available()` uses `Configured()` |
| embedding provider inference | never — uses `Configured()` |
| Telegram | `Run()` resolves; `NeedsSetup()` only checks configuration |
| web push | sending; auto-generation checks `Configured()` first |
| MCP headers / env / client secret | server connect |

So an `op://` key under `image.venice` does not prompt 1Password when you start a chat, and building the tool registry no longer touches a vault-backed `search.brave.api_key`.

Cascades keep their previous precedence (for example `audio.venice` → `image.venice`, each with its own `VENICE_API_KEY` fallback) through `ResolveFirstCredential`.

## Verification

`go build ./...` and `go test ./...` pass. New tests in `internal/config/credentials_test.go` cover literal/env/`${VAR}`/`$()`/`file://` resolution, that `Configured()` does not execute the deferred command, single-resolution memoization, cascade precedence, and that `Load()` resolves nothing.

Manual smoke test with a config whose `image.venice.api_key`, `search.brave.api_key` and `serve.telegram.token` were all `$(touch <marker>; printf key)`:

- `term-llm config get image.venice.api_key` → printed the raw `$()` string, **no marker created**
- `term-llm image "a cat"` → only the image marker appeared; Venice returned 401 for the fake key
- `term-llm notify telegram --chat-id 1 hi` → only the telegram marker appeared; Telegram returned 404 for the fake token
- the search marker was never created

One local test failure is unrelated to this change: `internal/serveui` `TestProductionBundleSizeBudgets` fails on my locally built frontend bundle (gzip 145656 vs budget 142000).


sam-saffron-jarvis:
CI status on head `e49dc34`: `changes`, `race`, `cross-build` and `passkey-smoke` pass.

`test` and `frontend` fail, and both also fail on the base commit `c6619d6a` on `main`, so neither is caused by this PR:

- `test` here failed only on `TestCollaborativeShellRealPTYStatePersistence` (a PTY timing test — "canceled shared shell command could not restore synchronization"). It passes locally with `-count=2`. The same job on `main@c6619d6a` fails on a different flaky test, `TestStreamReply_UncooperativeCleanupIsBounded`.
- `frontend` fails on `main@c6619d6a` too.

Happy to rerun the jobs if you want a cleaner board.

PR #1127, opened 2026-09-11
UX: Identify the sending node in completion push notifications

keegangeorge:
**Previously**, every node sent byte-identical push notification text, so anyone running a Hub with more than one subscribed node could not tell which conversation had finished without opening the notification.

**In this update**, the notification title is prefixed with the node's configured label (`uiTitle`, then `hubNodeName`, then `hubNodeID`), leaving deployments with no node identity on the original wording.

---

### Why

Push subscriptions are stored per node, so with a Hub fronting several serves each one pushes independently. The payload already carries a node-specific `url`, so tapping the notification lands in the right place, but the visible `title` and `body` are constants:

```go
title := "Response complete"
body  := "Your term-llm response is ready."
```

Two nodes finishing work produce indistinguishable notifications.

### Approach

`completionPushLabel()` resolves an identity from config the operator has already set, and returns `""` when none is configured so single-node deployments are unchanged.

`agentName` is deliberately **not** consulted: nearly every deployment passes `--agent`, so using it would change the wording for almost everyone rather than only for people who have actually labelled a node.

No frontend change is needed. `sw.js` already renders `raw.title` from the payload.

### Known gap

`frontend/src/platform/notifications.ts` builds the same payload shape for the same-device hidden-tab fallback and hardcodes the title too. Labelling that path needs the node label exposed to the browser (there is no `window.TERM_LLM_UI_TITLE` today), which felt like a separate change with a design decision attached. Happy to follow up if you want it, or to fold it in here if you would rather it land together.

### Testing

`TestCompletionPushTitleIdentifiesNode` asserts the enqueued payload rather than the helper, covering precedence, whitespace-only values falling through, failure outcomes, and the unlabelled case staying byte-identical.

```
ok  github.com/samsaffron/term-llm/cmd
```

### DIFF
--- a/cmd/serve_completion_push.go
+++ b/cmd/serve_completion_push.go
@@ -60,6 +60,26 @@ func (s *serveServer) validateCompletionPushTarget(ctx context.Context, id strin
 	return sub.ID, nil
 }
 
+// completionPushLabel returns a short label identifying the node that produced
+// a completion notification, or "" when this deployment has no node identity
+// configured.
+//
+// A Hub fronts several nodes and every node sends byte-identical notification
+// text, so once more than one node has a push subscription the notifications
+// are indistinguishable until you open one. The payload already carries a
+// node-specific URL, but the visible title does not.
+//
+// agentName is deliberately not consulted: nearly every deployment passes
+// --agent, and single-node users should keep the original wording.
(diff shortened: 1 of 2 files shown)


SamSaffron:
looks good, thanks Keegan

PR #1129, opened 2026-09-12
Use default sampling for Guardian reviews

supercobra:
Guardian review requests always sent `temperature: 0`, causing models that reject temperature control to fail tool approvals. Use the model's default sampling settings so these reviews can proceed.

Adds a regression test covering initial and follow-up reviews, checking that temperature is omitted and allow/deny decisions are preserved. The test failed before the fix and passes after it.

Validation:
- `gofmt` and `git diff --check` passed.
- `make build` and `go vet ./...` passed.
- `go test ./internal/guardian` passed; the full run also passed `internal/llm` and `internal/tools`.
- `go test ./...` was run with an isolated home and API-key environment. It failed outside the changed package: response timeout (`cmd`), hook timeout (`internal/gitcommit`), tmux capability (`internal/image`), bundle gzip size budget (`internal/serveui`), and ZDOTDIR handling (`internal/tui/chat`). These failures were not fixed in this PR.

Error was using GPT 5.6 sol and luna-max. Nemotron has the same issue if I remember well (not tested with this change).

No errors after this patch.

### DIFF
--- a/internal/guardian/guardian.go
+++ b/internal/guardian/guardian.go
@@ -362,7 +362,8 @@ func (r *Reviewer) turnMessages(policy string, req Request, mode PromptMode) []l
 }
 
 func (r *Reviewer) runReviewRequest(ctx context.Context, messages []llm.Message) (string, llm.Usage, error) {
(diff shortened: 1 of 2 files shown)


SamSaffron:
tricky one, I wanted 0 to give a more predictable guardian surface... but yeah this is a curveball for sure

PR #1153, opened 2026-09-19
Add built-in video editor agent

ducks:
People can now edit local video and audio through a purpose-built `@video-editor` agent instead of translating ordinary editing requests into FFmpeg commands themselves.

The agent inspects inputs with `ffprobe`, proposes cuts and transformations, creates a low-resolution preview when useful, renders with FFmpeg, and verifies every completed file. It keeps source media unchanged, writes temporary and final files under `work/` and `output/`, refuses silent overwrites, and requires user approval for commands that create or change files. Visual review uses selected frames rather than sending full videos to the model, and transcription is available when it helps choose an edit.

Validation:

- `make build`
- `go test ./internal/agents`
- `go vet -p 2 ./...`
- `./term-llm agents show video-editor`
- FFmpeg smoke test using a synthetic H.264/AAC source, preview render, and output probe


SamSaffron:
nice yeah sounds like a good idea

PR #1154, opened 2026-09-19
Update Go complexity baseline

ducks:
The hosted delegation work in `b1f90941` changed several existing functions and added four functions above the complexity threshold, but its generated complexity baseline was not refreshed. As a result, `make complexity` currently fails on `main` and on unrelated pull requests.

This regenerates the baseline from current `main`, preserving the ratchet's existing ownership and milestone metadata while recording the current measurements.

Validation: `make complexity`


SamSaffron:
thanks, addressed the complexity issues

PR #1167, opened 2026-09-21
feat: add bulk clearing for Hub review notifications

arpitjalan:
Screenshot 2026-09-21 at 21 39 15

Add a Clear all action to the Ready to review panel so completed notifications can be dismissed without opening each conversation. Clear the full unread list, including entries beyond the dashboard display limit, using the existing node acknowledgement API.

Acknowledge exact store identities and completion sequences to preserve newer completions. Coordinate clearing with collection to prevent stale notifications from returning, refresh counts, and report partial failures for retry. Preserve conversation history and pending questions or approvals.

Add regression coverage for durable acknowledgements, duplicate node registrations, authentication, partial failures, and stale polling. Verify the dashboard action on desktop and mobile layouts.


SamSaffron:
thanks Arpit give me a tiny bit, its a big change ... want to be careful with it.

